PR #15, opened 2025-11-19
[AAP-57100] Normalize credentials structure in dynamic job templates

abhikdps:
## Description

Launching a template from Self-Service fails if the template contains pre-populated multiple credentials.
This fix adds credential type, summary_field fields in dynamically generated AAP templates which makes sure all the required credential structure keys are present so that the execution does not fails when the `launch-job-template` action is run.
Also adds a logging line to output if multiple credentials of the same type are added.

## Related Issues

[AAP-57100](https://issues.redhat.com/browse/AAP-57100)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.test.ts
+++ b/plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.test.ts
@@ -230,6 +230,14 @@ const MOCK_JOB_TEMPLATE: IJobTemplate = {
       description: 'Test webhook credential',
       kind: 'github_token',
       cloud: false,
+      credential_type: 1,
+      summary_fields: {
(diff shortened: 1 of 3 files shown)


plugins/backstage-rhaap-common/src/AAPClient/AAPClient.ts
```diff
@@ -652,20 +652,32 @@ export class AAPClient implements IAAPService {
       const seen = new Set();
       const duplicates = [] as string[];
       payload.credentials.some(currentObject => {
```
audgirka: @abhikdps Are you sure we want to use `.some()` here?
`some()` will short circuit if it gets something as `true` and will not run for other items in the array

PR #62, opened 2025-12-24
[EE Builder Exp][AAP-61060] ensure description in generated software template is valid YAML

NilashishC:
## Description

Wrap the description field's value with `"` and replace `"` in the description with `'`.

## Related Issues

Closes https://issues.redhat.com/browse/AAP-61060

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Added/updated test cases.

N/A

## Checklist

- [x] Tests pass locally
- [x] Documentation updated

### DIFF
--- a/plugins/scaffolder-backend-module-backstage-rhaap/src/actions/createEEDefinition.ts
+++ b/plugins/scaffolder-backend-module-backstage-rhaap/src/actions/createEEDefinition.ts
@@ -879,7 +879,7 @@ kind: Template
 metadata:
   name: ${values.eeFileName}
   title: ${values.eeFileName}
-  description: "${(values.eeDescription || 'Saved Ansible Execution Environment Definition template').replaceAll('"', "'")}"
+  description: ${yaml.dump(values.eeDescription || 'Saved Ansible Execution Environment Definition template', { quotingType: '"', forceQuotes: true }).trim()}
   annotations:
     ansible.io/saved-template: 'true'
   tags: ${tagsJson}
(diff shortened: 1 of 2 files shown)


plugins/scaffolder-backend-module-backstage-rhaap/src/actions/createEEDefinition.ts
```diff
@@ -879,7 +879,7 @@ kind: Template
 metadata:
   name: ${values.eeFileName}
   title: ${values.eeFileName}
-  description: ${values.eeDescription || 'Saved Ansible Execution Environment Definition template'}
+  description: "${(values.eeDescription || 'Saved Ansible Execution Environment Definition template').replace(/"/g, "'")}"
```
audgirka: Can we use `js-yaml` as it will do the heavy lifting of parsing other special characters and adding escape sequences etc. 
Something like:
`yaml.dump('Line1\nLine2').trim()  // → |-\n  Line1\n  Line2 (block scalar)`

abhikdps:
```suggestion
  description: "${(values.eeDescription || 'Saved Ansible Execution Environment Definition template').replaceAll('"', "'")}"
```

NilashishC: From further testing, with `yaml.dump()` on `description`, a value such as the following isn't converted into valid YAML.

```
EE description with "double quotes" and 'single quotes' with : and other new chars
```

Generated template:
```yaml
---
apiVersion: scaffolder.backstage.io/v1beta3
kind: Template
metadata:
  name: ee
  title: ee
  description: >-
  EE description with "double quotes" and 'single quotes' with : and other new
  chars
  annotations:
    ansible.io/saved-template: 'true'
  tags: ["execution-environment"]
spec:
  type: execution-environment
  
< ... redacted for brevity ... >
```

PR #150, opened 2026-01-29
[AAP-63226][AAPClient] Add support for PAH API endpoints

NilashishC:
## Description

Adds support for discovering collections from Private Automation Hub in AAPClient.

## Related Issues

Closes https://issues.redhat.com/browse/AAP-63226

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Test cases added.

N/A

## Checklist

- [x] Tests pass locally
- [x] Documentation updated

### DIFF
--- a/plugins/self-service/src/components/CreateTask/StepForm.test.tsx
+++ b/plugins/self-service/src/components/CreateTask/StepForm.test.tsx
@@ -838,6 +838,117 @@ describe('StepForm', () => {
     });
   });
 
+  describe('allOf and nested object support', () => {
+    it('extracts uiSchema from allOf and nested objects, and includes allOf properties in review', async () => {
+      let capturedUiSchema: any;
+      jest
+        .spyOn(require('./ScaffolderFormWrapper'), 'ScaffolderForm')
+        .mockImplementation(({ uiSchema, onSubmit, children }: any) => {
+          capturedUiSchema = uiSchema;
+          return (
+            <form
+              onSubmit={e => {
+                e.preventDefault();
+                onSubmit({
+                  formData: { name: 'test', conditionalField: 'value' },
+                });
+              }}
+            >
+              <div>MockForm</div>
+              {children}
+              <button type="submit">Submit</button>
+            </form>
+          );
+        });
+
+      const fieldOrder = ['name', 'config', 'conditionalField', '*'];
+      const steps = [
+        {
+          title: 'Step',
+          schema: {
+            'ui:order': fieldOrder,
+            properties: {
+              name: { type: 'string', title: 'Original' },
+              emptyObj: { type: 'object' },
+              config: {
+                type: 'object',
+                'ui:title': 'Config',
+                properties: {
+                  flag: { type: 'boolean', 'ui:help': 'Toggle' },
+                },
+                dependencies: {
+                  flag: {
+                    oneOf: [
+                      {
+                        properties: {
+                          flag: { const: true },
+                          depField: {
+                            type: 'string',
+                            'ui:field': 'DepPicker',
+                          },
+                        },
+                      },
+                    ],
+                  },
+                },
+                allOf: [
+                  {
+                    then: {
+                      properties: {
+                        nestedAllOf: {
+                          type: 'string',
+                          'ui:field': 'AllOfPicker',
+                        },
+                      },
+                    },
+                  },
+                ],
+              },
+            },
+            allOf: [
+              {
+                then: {
+                  properties: {
+                    conditionalField: {
+                      type: 'string',
+                      title: 'Conditional',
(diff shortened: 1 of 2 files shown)


plugins/backstage-rhaap-common/src/interfaces/Collection.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Represents a collection version returned by the Private Automation Hub (PAH)
+ * collection-versions search API, as normalized by getCollectionsByRepositories.
+ */
+export interface ICollection {
```
NilashishC: We can drop the `I` from the interface name here. Looks like it was an old convention.

plugins/backstage-rhaap-common/src/AAPClient/pahHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+): repositories is string[] {
+  if (!repositories || !Array.isArray(repositories)) {
+    context.logger.warn(
+      `[${context.pluginLogName}]: Invalid repositories parameter. Expected an array, received: ${typeof repositories}`,
+    );
+    return false;
+  }
+  if (repositories.length === 0) {
+    context.logger.info(
+      `[${context.pluginLogName}]: No repositories provided. Returning empty collection list.`,
```
audgirka: Since this function is returning a boolean, maybe this message make more sense in the `syncCollectionsByRepositories()`

plugins/backstage-rhaap-common/src/AAPClient/pahHelpers.ts
```diff
@@ -0,0 +1,210 @@
...
+  logger: LoggerService;
+  pluginLogName: string;
+  executeGetRequest: (url: string, token: string | null) => Promise<Response>;
+  isValidPAHRepository: (repositoryName: string) => Promise<boolean>;
+}
+
+export function validateRepositoriesInput(
+  repositories: unknown,
+  context: PAHHelperContext,
+): repositories is string[] {
```
audgirka: This function is returning a `boolean` instead of `string[]`

plugins/backstage-rhaap-common/src/AAPClient/pahHelpers.ts
PyContriver: That is a good point, made the changes to move the log message to `syncCollectionsByRepositories()` in the commits [956f81a](https://github.com/ansible/ansible-backstage-plugins/pull/150/commits/956f81a537ec8630bfbb4225a5cd6f923434a8d3) and [57679c3](https://github.com/ansible/ansible-backstage-plugins/pull/150/commits/57679c3d670b548fa7de49f4e14b545f76e63ed3)

plugins/backstage-rhaap-common/src/AAPClient/pahHelpers.ts
PyContriver: @audgirka So This is a TypeScript type predicate and not a return type of string[]. it considers:

- At runtime: Returns a boolean (true or false)
- At compile time: Tells TypeScript that if the function returns true, the repositories parameter can be safely treated as string[]. 

I did try to add return type as well as boolean but it results in syntax error: 

`boolean & (repositories is string[])`
 
I have added a JsDoc comment here [b53fbcc](https://github.com/ansible/ansible-backstage-plugins/pull/150/commits/b53fbccebc18cdea34338966f240e7210c8a06e5) to avoid any confusion.

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.ts
```diff
@@ -1324,4 +1343,85 @@ export class AAPClient implements IAAPService {
...
+    const token = this.ansibleConfig.rhaap?.token ?? null;
+
+    const context: PAHHelperContext = {
+      logger: this.logger,
+      pluginLogName: this.pluginLogName,
+      executeGetRequest: this.executeGetRequest.bind(this),
+      isValidPAHRepository: this.isValidPAHRepository.bind(this),
+    };
+
+    const isEmptyArray =
```
audgirka: Why an additional check the param on line 1359 already tells us this is a `string[]`

plugins/backstage-rhaap-common/src/AAPClient/pahHelpers.ts
PyContriver: removed this and moved the logic to `syncCollectionsByRepositories` directly in [5447a4b](https://github.com/ansible/ansible-backstage-plugins/pull/150/commits/5447a4ba7e6f0adefe5572ce34b9d9fd514a0ed5)

PyContriver (approved):

PR #145, opened 2026-01-27
fix: Execution Environments catalog creation order

Pavanjangle:
## Description

Added client-side sorting for catalog items by creation order

## Related Issues

[AAP-63831
](https://issues.redhat.com/browse/AAP-63831)
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Code follows project style
- [x] Tests pass locally
- [ ] Documentation updated

PR #157, opened 2026-02-03
feat: Add custom EETagsPicker component for EE template tags

Pavanjangle:
## Description

Implemented a custom `EETagsPicker` UI component for Execution Environment templates, allowing users to add, remove, and reorder tags with built-in validation based on Backstage rules.

## Related Issues

[AAP-63487](https://issues.redhat.com/browse/AAP-63487)

## Template PR Link

https://github.com/ansible/ansible-rhdh-templates/pull/45

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

### DIFF
--- a/plugins/self-service/src/components/ExecutionEnvironments/catalog/EEDetailsPage.tsx
+++ b/plugins/self-service/src/components/ExecutionEnvironments/catalog/EEDetailsPage.tsx
@@ -31,7 +31,7 @@ export const EEDetailsPage: React.FC = () => {
   };
   const handleMenuClose = () => setAnchorEl(null);
   const catalogApi = useApi(catalogApiRef);
-  const [entity, setEntity] = useState<any | null>(null);
+  const [entity, setEntity] = useState<any | null>(false);
   const [menuid, setMenuId] = useState<string>('');
   const [defaultReadme, setDefaultReadme] = useState<string>('');
   const discoveryApi = useApi(discoveryApiRef);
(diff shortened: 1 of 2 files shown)


plugins/self-service/src/components/ExecutionEnvironments/catalog/EEDetailsPage.tsx
NilashishC: Could you please explain the changes in this file? Why are we updating it as a part of this work?

plugins/self-service/src/utils/validationsUtils.ts
```diff
@@ -0,0 +1,66 @@
+export const isValidEntityName = (
```
NilashishC: Is there a strong reason of using `isValidEntityName()` to support tag validation? A `tag` isn't really an entity. And the `isTag` param contradicts the name of the function.

NilashishC (changes requested):
I highly recommend not touching existing components as a part of this PR. The validations can stay in the individual component files for now.

plugins/self-service/src/components/ExecutionEnvironments/catalog/EEDetailsPage.tsx
abhikdps: this is something that went as part of PR #145 . i think we decided to just revert this change. This is not a newly introduced change. @Pavanjangle please start a new PR for the same if required.

NilashishC (approved):

PR #185, opened 2026-02-26
refactor: Execution Environment details page layout

Pavanjangle:
## Description

- Updated the Execution Environment details page to match the Figma design. The README is now shown on the left, and the right side has a sidebar with About, Defined Content, Resources, and optional Links.

- Replaced the kebab menu with an Actions dropdown that includes Build, Edit definition, View in source, and Delete. The Edit definition option now opens the EE definition YAML file.

- Added new Defined Content and Resources cards. Moved the definition parsing logic to eeDefinitionUtils so we can display the base image, collections, Python packages, and system packages from the definition.

## Related Issues

[AAP-66054](https://issues.redhat.com/browse/AAP-66054)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

### DIFF
--- a/plugins/self-service/src/components/ExecutionEnvironments/catalog/AboutCard.tsx
+++ b/plugins/self-service/src/components/ExecutionEnvironments/catalog/AboutCard.tsx
@@ -153,7 +153,6 @@ export const AboutCard: React.FC<AboutCardProps> = ({
             </Box>
           )}
         </Box>
-        <Divider style={{ margin: '12px -16px 12px' }} />
 
         {/* Description - inline expand when > 150 chars, no TechDocs link */}
         <Box>
--- a/plugins/self-service/src/components/ExecutionEnvironments/catalog/DefinedContentCard.tsx
+++ b/plugins/self-service/src/components/ExecutionEnvironments/catalog/DefinedContentCard.tsx
@@ -1,22 +1,49 @@
-import { Box, Card, CardContent, Typography } from '@material-ui/core';
+import { Box, Card, CardContent, Typography, Divider } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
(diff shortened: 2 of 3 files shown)


NilashishC:
> @NilashishC PR #194 has breakingn changes which will affect this

I've resolved them.

plugins/self-service/src/components/ExecutionEnvironments/catalog/EEDetailsPage.tsx
```diff
@@ -193,6 +292,43 @@ export const EEDetailsPage: React.FC = () => {
...
 
+  /** URL to edit the EE definition file (e.g. test-2.yaml), not catalog-info.yaml */
+  const getDefinitionEditUrl = useCallback(() => {
+    const editUrl = entity?.metadata?.annotations?.[ANNOTATION_EDIT_URL] as
+      | string
+      | undefined;
+    if (!editUrl) return null;
+    const definitionFilename =
+      entity?.metadata?.annotations?.['ansible.io/ee-definition-filename'] ??
+      `${entity?.metadata?.name ?? 'execution-environment'}.yaml`;
```
abhikdps: Instead of using `execution-environment`, we should throw an error, if the filename is not found

NilashishC: This scenario won't theoritically be possible coz entities cannot be created without names. But I'll update this to make sure it errors out anyway. Also, the `ansible.io/ee-definition-filename` isn't set anywhere, unsure why it was added in the first place. I'll remove it and only look for `entity?.metadata?.name`.

PR #189, opened 2026-03-04
Adds `EEEntityRegistrar` to replace `/register_ee` API

NilashishC:
### Description

- Introduce a root-scoped `EEEntityRegistrar` service ref in `backstage-rhaap-common` so the catalog and scaffolder plugins share a single instance for registering Execution Environment entities, eliminating the need for an HTTP round-trip between them.

- Remove the `/register_ee` endpoint from the catalog router, along with all associated tests and the auth/discovery dependencies that were only needed to call it.

- Simplify EEEntityProvider to a thin adapter that bridges the shared registrar with the catalog's EntityProviderConnection.

**Motivation**

- Previously, when a user created an EE definition without publishing to SCM, the scaffolder action (createEEDefinitionAction) had to make an **authenticated** HTTP POST to the catalog plugin's `/register_ee` endpoint to register the entity. This introduced unnecessary coupling: the scaffolder needed `AuthService` and `DiscoveryService` solely to construct a fetch call to another plugin running in the same backend process.

- By using a root-scoped Backstage service ref (`eeEntityRegistrarRef`), both plugins receive the same `EEEntityRegistrar` instance. The catalog module wires it to the `EntityProviderConnection` via `setConnection`, and the scaffolder module calls registerExecutionEnvironment directly.

## Related Issues

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tests added/updated.

N/A

## Checklist

- [x] Tests pass locally
- [x] Documentation updated


audgirka (approved):
looks good, but would need more testing for this

audgirka:
@NilashishC can we close this PR?

NilashishC:
Closing this since the main security concern was resolved by only allowing service-to-service credentials.

PR #195, opened 2026-03-12
Fix: portal - favorite entities return 404 error

ankitsharma589:
## Description

**Issue:** Favorite entities were returning a 404 error on the portal because the browser was calling `/api/user-settings/buckets/default.starredEntities/keys/entityRefs`, but no backend existed to handle that route.

**Root cause:** The `@backstage/plugin-user-settings-backend` was not installed. In local dev, starring worked only because it used browser localStorage (no backend call needed), masking the problem.

**Fix :**
1. Added `@backstage/plugin-user-settings-backend` to backend `package.json`
2. Registered it in backend `index.ts` via `backend.add(...)`
3. Overrode `storageApiRef` in the app's `apis.ts` to use `UserSettingsStorage` instead of the default `WebStorage`, so the frontend stores starred entities via the backend API instead of localStorage

## Related Issues

Closes [AAP-66227](https://issues.redhat.com/browse/AAP-66227)

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing
All local tests passed. No testcase added.

**Before fix:**

- Chrome

Screenshot From 2026-03-12 20-13-42

- Firefox

Screenshot From 2026-03-12 20-09-47

- Curl Output on API calling:

Screenshot From 2026-03-12 20-28-48

**After fix:**

- Chrome

Screenshot From 2026-03-12 20-14-46

- Firefox

Screenshot From 2026-03-12 20-13-42

- Curl Output on API calling:

image
image

## Checklist

- [X] Code follows project style
- [X] Tests pass locally
- [ ] Documentation updated


audgirka:
not reproducible on rhel installer

PR #213, opened 2026-03-30
Fix job templates not appearing after sync without page refresh

Jeevansuresh1:
## Description

Fixes an issue where newly synced Job Templates were not visible unless the page was manually refreshed.

After sync:
- Refreshes the jobTemplates allowlist from AAP  
- Remounts EntityListProvider to trigger catalog reload  
- Reload happens only when actual changes are detected (ID comparison)

This ensures new templates appear automatically without requiring user action.

## Related Issues

Closes AAP-52257

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- Synced new Job Templates and verified they appear immediately  
- Verified no reload when there are no changes  
- Tested add/remove/update scenarios  

## Checklist

- [x] Tests pass locally  
- [ ] Documentation updated  

## Summary by CodeRabbit

* **Tests**
  * Added coverage for fetching job templates, the "Sync now" flow, and refresh behavior—verifying re-fetches after successful sync, no redundant re-fetch on failure, and remounts when template sets change.

* **Refactor**
  * Improved template fetch/sync flow to prevent stale updates, show intermediate status, conditionally retry fetches, and ensure the template list remounts for fresh data (including on navigation).


NilashishC (changes requested):
Create a new template -> Sync -> It should appear in the catalog -> Rename the same template -> Sync. This time the template does not update and requires a page refresh to get the updated name.

plugins/self-service/src/components/RouteView/RouteView.tsx
```diff
@@ -47,7 +48,7 @@ const RouteViewContent = () => {
   return (
     <>
       <Routes>
-        <Route path="catalog" element={<HomeComponent />} />
+        <Route path="catalog" element={<HomeComponent key={location.key} />} />
```
NilashishC: Using `location.key` as the `key` prop means `HomeComponent` fully remounts on every navigation to this route — destroying all state including user filter selections, scroll position, in-flight API calls, and `EntityListProvider` cached entity data.

The `syncKey` mechanism inside `Home.tsx` already handles sync-triggered refreshes surgically. This `location.key` remount is a separate, heavier approach that is not mentioned in the PR description (which says "Reload happens only when actual changes are detected").

Could you clarify the intent here? If the goal is to ensure fresh data when navigating to the templates page from another route, a targeted refetch (e.g., calling `fetchJobTemplates()` inside a `useEffect` keyed to a navigation signal) would preserve the user's filter state instead of tearing down the whole tree.

NilashishC (changes requested):
The core approach is solid — centralizing `fetchJobTemplates`, comparing pre/post template sets with `id:name` serialization, and using `syncKey` to remount `EntityListProvider` is a clean solution. The `fetchRequestIdRef` stale-request guard and `jobTemplatesRef` patterns are well done. Good test coverage at 97.7%.

Two things to address before this is ready:

1. **Success snackbar on fetch failure** — line 265 shows "Templates synced successfully" even when `fetchJobTemplates()` returns `undefined`. This was raised by CodeRabbit three times. See inline comment for a suggested fix.

2. **`key={location.key}` on `HomeComponent`** — this causes a full state teardown on every navigation, which seems heavier than intended. Would like to understand the use case or see a lighter alternative.

See inline comments for details.

plugins/self-service/src/components/Home/Home.test.tsx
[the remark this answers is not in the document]
Jeevansuresh1: Resolved in my recent commit.

NilashishC (approved):
All three issues from my previous review have been addressed:

1. **Snackbar fix** — conditional message on lines 267-271 properly distinguishes success from fetch failure. Error logging added in catch block.
2. **Tests** — now verify actual `EntityListProvider` remount via `getEntityFacets` call counts, not just `autocomplete` calls.
3. **`key={location.key}`** — your explanation about the catalog-import navigation flow makes sense. `syncKey` only covers the sync button; this covers the navigation case.

Re: the rename scenario I flagged separately — the comparison logic (`${id}:${name}` serialization) correctly detects renames, and the test at line 643-679 proves it. If there was a manual-testing issue, it's likely backend timing (entity provider not finished before the frontend re-queries), which is outside this PR's scope.

Minor note: patch coverage dropped to 90% — the 4 uncovered lines are the error catch/snackbar-failure paths. Consider adding a test that mocks `getAccessToken` or `autocomplete` to throw, but not a blocker.

plugins/self-service/src/components/Home/Home.test.tsx
```diff
@@ -408,6 +408,291 @@ describe('self-service', () => {
...
+        },
+        { timeout: 4000 },
+      );
+
+      await waitFor(() => {
+        expect(
+          mockCatalogApi.getEntityFacets.mock.calls.length,
+        ).toBeGreaterThan(facetCallsBeforeSync);
+      });
+    });
```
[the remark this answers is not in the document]
Jeevansuresh1: This test is correct and intentional. Looking at the handleSync implementation in Home.tsx, setSyncKey(prev => prev + 1) is called unconditionally after a successful sync — regardless of whether the AAP job template list changed.

plugins/self-service/src/components/Home/Home.tsx
```diff
@@ -227,27 +302,22 @@ export const HomeComponent = () => {
...
-  }, [scaffolderApi, rhAapAuthApi]);
+    fetchJobTemplates();
+  }, [fetchJobTemplates]);
+
+  // After fetchJobTemplates completes, schedule a catalog refresh so that
+  // recently imported templates (via "Add Template") have time to be
+  // processed by the catalog backend before we re-query.
+  useEffect(() => {
+    if (loading) return undefined;
+    const CATALOG_SETTLE_MS = 3000;
```
abhikdps:
```suggestion
    const CATALOG_SETTLE_MS = 750;
```

abhikdps: make this as low as possible. i tested with 750 and it works works. going lower than this does not refresh with the latest entities.

abhikdps (approved):

audgirka (approved):

PR #250, opened 2026-04-15
Update import to fix deployment with rhdh-1.8

abhikdps:
## Description

Update import to fix deployment with rhdh-1.8

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tested with rhdh-1.8 and rhdh-1.9

## Summary by CodeRabbit

* **Chores**
  * Updated plugin scaffolder dependency constraints to support a broader range of compatible versions, improving installation flexibility.
  * Refactored internal module resolution logic for enhanced compatibility across dependency versions.

PR #255, opened 2026-04-17
fix: use direct named import for scaffolderActionsExtensionPoint

ganeshrn:
## Summary

- Fix scaffolder plugin crash on startup that blocks OAuth login in QCOW2 appliance

## Problem

The namespace import (`import * as scaffolderNode`) introduced in b6fa6e8 does not propagate named exports through the dynamic plugin bundler's CommonJS interop at runtime. This causes `scaffolderActionsExtensionPoint` to resolve to `undefined`, crashing the scaffolder plugin on startup:

```
Plugin 'scaffolder' threw an error during startup
TypeError: Cannot read properties of undefined (reading 'id')
    at BackendInitializer.cjs.js:262:72
```

This blocks OAuth login because the catalog cannot index user entities when the scaffolder plugin fails to initialize.

## Fix

Revert to the direct named import which resolves correctly in the dynamic plugin context:

```typescript
import { scaffolderActionsExtensionPoint } from '@backstage/plugin-scaffolder-node';
```

RHDH 1.8 ships `@backstage/plugin-scaffolder-node@0.12.1` which exports `scaffolderActionsExtensionPoint` from the stable path.

## Test plan

- [x] Deploy pre-built QCOW2 appliance on KVM with cloud-init
- [x] Verify scaffolder plugin crash present without fix
- [x] Apply fix (patch bundled `module.cjs.js` on VM)
- [x] Verify no scaffolder errors after restart
- [x] Verify OAuth login succeeds after fix

## Jira

Fixes: https://issues.redhat.com/browse/AAP-72260
Parent: https://issues.redhat.com/browse/ANSTRAT-1897

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal module dependencies with no user-facing changes.

### DIFF
--- a/plugins/scaffolder-backend-module-backstage-rhaap/src/module.ts
+++ b/plugins/scaffolder-backend-module-backstage-rhaap/src/module.ts
@@ -18,14 +18,11 @@ import {
   coreServices,
   createBackendModule,
 } from '@backstage/backend-plugin-api';
-import type { ExtensionPoint } from '@backstage/backend-plugin-api';
-import type { ScaffolderActionsExtensionPoint } from '@backstage/plugin-scaffolder-node';
-import * as scaffolderNode from '@backstage/plugin-scaffolder-node';
+import { scaffolderActionsExtensionPoint } from '@backstage/plugin-scaffolder-node';
 import {
   scaffolderAutocompleteExtensionPoint,
   scaffolderTemplatingExtensionPoint,
 } from '@backstage/plugin-scaffolder-node/alpha';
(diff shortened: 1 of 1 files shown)


audgirka:
@abhikdps this is already covered as part of 1.9 upgrade correct?

abhikdps:
@audgirka This reverts the change in #250 . The issue was that it was not working for me with rhdh-1.8. But if RHDH 1.8 does ships `@backstage/plugin-scaffolder-node@0.12.1`, then it should work with 1.8 as well like @ganeshrn mentioned.

abhikdps:
@ganeshrn This change should be merged to the `release-2.1` branch and not to main.

PR #267, opened 2026-04-22
fix(scm): refresh GitHub installation tokens on each request

NilashishC:
## Description

GithubClient previously froze the Bearer string from ScmClientFactory at construction time, so GitHub App installation tokens expired after ~1h and catalog sync failed with Bad credentials. Add optional getToken on ScmClientConfig, resolve the bearer in getFetchOptions before each fetch, and have ScmClientFactory supply getToken when not using a user-provided token.

Extend unit tests to assert getToken is preferred across multiple requests. Import ScmClientConfig for JSDoc @link resolution.

### GitHub SCM credentials: token lifecycle and catalog sync

This document describes why long-running Backstage processes could see GitHub API failures (for example `Bad credentials`) when using a GitHub App or installation-backed integration, how that behavior showed up in this repository, and how it was addressed.

### Root cause

GitHub’s server-to-server authentication uses **short-lived** credentials:

1. **App JWT**: Signed with the app’s private key; GitHub accepts it only for a **short window** (on the order of minutes). It is used to obtain installation tokens, not as the long-term API credential for GraphQL/REST.
2. **Installation access token**: What you typically send as `Authorization: Bearer …` to the GitHub API. It expires after **about one hour**.

In this codebase, **`GithubClient`** originally set `Authorization` from a **single static string** (`config.token`) captured when the client was constructed. It did **not** refresh that value over time.

The **Ansible Git Contents** catalog path constructs **one** `GithubClient` (via `ScmClientFactory` → `ScmCrawlerFactory`) when the provider is initialized at backend startup, then reuses that client for **every scheduled sync**. The bearer token was therefore effectively **frozen for the lifetime of the process** (or until restart), even though GitHub invalidates installation tokens on a fixed TTL.

After enough time (or after a long single sync crossing the TTL boundary), GitHub rejects requests. Failures are often reported as **`Bad credentials`** or similar auth errors.

So the root cause was **credential lifetime mismatch**: GitHub tokens expire; the client kept using an expired token.

## How credentials flow from app-config to `Authorization: Bearer`

Backstage reads `integrations.github` from merged `app-config` into a **`ScmIntegrations`** registry. **`DefaultGithubCredentialsProvider`** (from `@backstage/integration`) is constructed from that registry and knows how to issue credentials for a repository or organization URL.

For a **GitHub App**–backed integration, the private key in config is **not** turned into the Bearer value your HTTP client sends for GraphQL/REST. The chain is:

1. **Private key → short-lived App JWT** — The PEM is used to sign a JWT that identifies the GitHub App to GitHub’s API. That JWT is accepted only for a brief period and is meant for **token exchange**, not for listing repos or calling the GraphQL API as the installation.
2. **App JWT → installation access token** — The credentials provider calls GitHub (e.g. to create an installation access token for the org/repo implied by the URL). GitHub returns a **Bearer** string valid for roughly an hour.
3. **This repository**: `resolveGithubToken` (used from `ScmClientFactory`) calls `credentialsProvider.getCredentials({ url })` and uses `credentials.token` as the string to authenticate. If that path fails, it falls back to the static PAT on the integration entry, `integrations.github[].token`.

So the value `GithubClient` places on `Authorization: Bearer` is either that **installation access token** (App path) or a **PAT** (fallback / PAT-only config)—never the raw private key.

## Before the fix in this PR: what actually went wrong

The bug was **not** that the wrong *kind* of credential was used for App-based setups. After a successful `getCredentials` at client construction time, requests already used a normal **installation access token**, which is the correct credential type for server-to-server API access.

The bug was that this token was **copied once** into `GithubClient`’s static `config.token` and **never renewed**, while GitHub continued to expire installation tokens on schedule. So after about an hour (or a long sync crossing that boundary), GitHub correctly rejected a **stale** installation token.

GithubClient does call `getToken()` (and thus `getCredentials`) before each HTTP request when that callback is set, so you re-run the resolution path every time.

That does not mean GitHub’s API mints a new installation token on every request. Backstage’s SingleInstanceGithubCredentialsProvider keeps a per-org in-memory Map (Cache.tokenCache) and only runs the code that calls createInstallationAccessToken when the cache is missing or the cached entry is past its effective expiry (they treat a token as stale 10 minutes before GitHub’s expires_at).

Ref: https://backstage.io/api/next/classes/_backstage_integration.DefaultGithubCredentialsProvider.html

Deployments that only use a **long-lived PAT** in `integrations.github[].token` (and never rely on the App path) may not hit hourly expiry; the failure mode is most visible when **`DefaultGithubCredentialsProvider`** supplies installation tokens.

### 1. New `GithubClient` instance each catalog sync

**Idea:** At the start of each `AnsibleGitContentsProvider` sync (or each crawler run), call `ScmClientFactory.createClient` again and use a new client for that sync only.

| Pros | Cons |
|------|------|
| Simple mental model: one token snapshot per sync | A **single** sync that runs **longer than the token TTL** can still fail **mid-sync** if the token is only obtained at sync start |
| Keeps `GithubClient` free of refresh logic | Extra object allocation each sync (usually minor) |
| Fewer calls into the credentials provider **during** a sync | Requires changes in the catalog provider / crawler wiring, not only in `GithubClient` |

### 2. Lazy token resolution on the existing client (chosen)

**Idea:** Keep one `GithubClient` instance, but resolve the bearer token **when issuing HTTP requests** (for example via an optional `getToken()` callback on `ScmClientConfig` that delegates to `resolveGithubToken` / `DefaultGithubCredentialsProvider`).

| Pros | Cons |
|------|------|
| Correct behavior for **long-running** work and for **many** scheduled syncs without restarting the backend | Slightly more complexity in `GithubClient` and config types |
| Does **not** require recreating the client each sync | More invocations of the credentials path **per HTTP request** (mitigated by provider-side caching) |
| One implementation benefits all call sites using `ScmClientFactory` for GitHub without `providedToken` | Needs clear documentation (this file) and tests |

### 3. Time-based in-client cache (hybrid)

**Idea:** Keep `getToken`, but cache the resolved string in `GithubClient` for e.g. 45–50 minutes and only call the provider again after TTL.

| Pros | Cons |
|------|------|
| Reduces calls to the credentials provider | Another TTL to tune and keep aligned with GitHub’s behavior |
| Still one long-lived client | More edge cases (clock skew, forced revocation) unless combined with retry-on-401 |

This was **not** implemented initially; it can be added later if profiling shows provider churn as a problem.

## Approach chosen

**Approach 2 — lazy token resolution** — was implemented:

- `ScmClientConfig` may include optional `getToken?: () => Promise<string | undefined>`.
- `GithubClient` prefers the result of `getToken()` for each request, then falls back to static `token` (unchanged behavior for PATs / explicit OAuth tokens).
- `ScmClientFactory` registers `getToken` for GitHub when **no** user `providedToken` is passed, so integration / GitHub App installation tokens are re-resolved through Backstage’s credentials provider on an ongoing basis.

This balances **correctness** (tokens can refresh during long syncs and across process uptime) with the requirement to **avoid recreating `GithubClient` on every sync**.

## Related Issues

Closes https://redhat.atlassian.net/browse/AAP-72657

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tests added / updated.

## Checklist

- [x] Tests pass locally
- [x] Documentation updated

## Summary by CodeRabbit

* **New Features**
  * GitHub auth can now refresh tokens per request when configured, while continuing to support static tokens.
* **Bug Fixes**
  * Failures during token resolution emit a warning and avoid sending stale/static tokens.
* **Documentation**
  * Configuration schema extended with an optional per-request token callback.
* **Tests**
  * Added coverage for per-request token usage and failure handling.


NilashishC:
Reasoning why coderabbitai's nitpick review on adding a TTL layer on top of getCredentials(), can be skipped:

**1. GitHub installation tokens are already de-duplicated in Backstage**
In @backstage/integration, SingleInstanceGithubCredentialsProvider uses an in-memory Map (Cache.tokenCache) and only calls the code path that does createInstallationAccessToken when the entry is missing or past an effective expiry (with a 10-minute buffer). So “every getCredentials” is not “every time hit GitHub for a new installation token.”

**2. The hot path is mostly “call in again,” not “re-mint from GitHub”**
It is right that `getToken` runs on every HTTP request in your GithubClient and that it goes through resolveGithubCredentials → resolveGithubToken → getCredentials. The expensive part (minting a token) is what Backstage already gates. The rest is mostly URL building, byHost lookup, try/catch, and logging, which indeed is real work, but usually not the same order of cost as a GitHub API token creation.

**3. Another TTL cache in your layer is easy to get wrong**
You’d be maintaining a second clock (your TTL vs Backstage’s buffer vs GitHub’s expires_at), extra tests, and a risk of serving a stale string if the cache key/TTL is wrong. That’s a lot of surface area for a problem Backstage is already trying to solve.

NilashishC:
@coderabbitai review

abhikdps (approved):

PR #277, opened 2026-04-24
Stop next button from triggering on error and make namespace and repo name mandatory

abhikdps:
## Description

- Make namespace and repo name mandatory in the SCMSelection extension.

Till commit [2a94e91](https://github.com/ansible/ansible-backstage-plugins/pull/277/commits/2a94e91df5e1a69e9f8b0ec20837bc48a37a0e82) the behaviour was:
- Disable next button on any error from the custom pickers in the Software template.

After coderabbit's comment https://github.com/ansible/ansible-backstage-plugins/pull/277#pullrequestreview-4173499482, the behaviour is changed to:
- Next button is always enabled and focusable. When `hasErrors` is `true`, `type="button"` stops it from triggering form submission, and `onClick` calls `notifySubmitAttempted` to reveal the field tooltips. When `hasErrors` is `false`, `type="submit"` submits the form normally. Keyboard and screen-reader users get the same experience as mouse users. This also makes it consistent with how backstage handles the next button.

## Related Issues

Closes [AAP-72626](https://redhat.atlassian.net/browse/AAP-72626) and [AAP-72628](https://redhat.atlassian.net/browse/AAP-72628)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tested locally.

Before https://github.com/ansible/ansible-backstage-plugins/pull/277#pullrequestreview-4173499482: 
Screenshot 2026-04-25 at 02 44 35

After https://github.com/ansible/ansible-backstage-plugins/pull/277#pullrequestreview-4173499482, the button is enabled but does not trigger incase of an error:
Screenshot 2026-04-25 at 03 12 53

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Release Notes

* **New Features**
  * Implemented centralized field-level validation system with real-time feedback across form components.
  * Submit button now prevents submission when required fields are invalid or empty.
  * Added inline error messages that display when attempting to submit incomplete forms.

* **Tests**
  * Added comprehensive validation behavior tests for form field components.

### DIFF
--- a/plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
+++ b/plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
@@ -1,5 +1,6 @@
 import { useState, useEffect, useMemo } from 'react';
 import { FieldExtensionComponentProps } from '@backstage/plugin-scaffolder-react';
+import { useFieldValidation } from '../../CreateTask/FieldValidationContext';
 import { TextField, Typography, Box, IconButton } from '@material-ui/core';
 import { Alert } from '@material-ui/lab';
 import { makeStyles } from '@material-ui/core/styles';
@@ -140,6 +141,7 @@ export const EETagsPickerExtension = ({
   rawErrors = [],
   schema,
   formData,
+  idSchema,
 }: FieldExtensionComponentProps<string[]>) => {
   const classes = useStyles();
   const requiredField = required ? required : true;
@@ -151,6 +153,17 @@ export const EETagsPickerExtension = ({
   const customDescription = schema?.description;
   const tags = formData ?? [];
   const [tagErrors, setTagErrors] = useState<Record<number, string>>({});
+
+  const { setFieldError } = useFieldValidation();
+  const fieldId = idSchema.$id;
+
+  useEffect(() => {
+    const hasTagError = Object.keys(tagErrors).length > 0;
+    setFieldError(fieldId, hasTagError);
+    return () => {
+      setFieldError(fieldId, false);
+    };
+  }, [fieldId, tagErrors, setFieldError]);
 
   const validateTag = (tag: string, index: number): boolean => {
     const validation = isValidTag(tag);
--- a/plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
+++ b/plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
@@ -158,7 +158,7 @@ export const EETagsPickerExtension = ({
   const fieldId = idSchema.$id;
 
   useEffect(() => {
-    const hasTagError = Object.keys(tagErrors).length > 0;
+    const hasTagError = Object.values(tagErrors).some(Boolean);
     setFieldError(fieldId, hasTagError);
     return () => {
       setFieldError(fieldId, false);
@@ -187,7 +187,11 @@ export const EETagsPickerExtension = ({
 
   const handleTagChange = (index: number, value: string) => {
     if (value.length === 0) {
-      setTagErrors(prev => ({ ...prev, [index]: '' }));
+      setTagErrors(prev => {
+        const next = { ...prev };
+        delete next[index];
+        return next;
+      });
     } else {
       validateTag(value, index);
     }
--- a/plugins/self-service/src/components/Scaffolder/ScmSelector/ScmSelectorExtension.tsx
+++ b/plugins/self-service/src/components/Scaffolder/ScmSelector/ScmSelectorExtension.tsx
@@ -21,6 +21,7 @@ import { makeStyles } from '@material-ui/core/styles';
 import { Alert } from '@material-ui/lab';
 import CheckCircleIcon from '@material-ui/icons/CheckCircle';
 import ErrorIcon from '@material-ui/icons/Error';
+import { useFieldValidation } from '../../CreateTask/FieldValidationContext';
 
 type ProviderType = 'github' | 'gitlab';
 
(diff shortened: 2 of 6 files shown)


plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
```diff
@@ -152,6 +154,17 @@ export const EETagsPickerExtension = ({
   const tags = formData ?? [];
   const [tagErrors, setTagErrors] = useState<Record<number, string>>({});
 
+  const { setFieldError } = useFieldValidation();
+  const fieldId = idSchema.$id;
+
+  useEffect(() => {
+    const hasTagError = Object.keys(tagErrors).length > 0;
```
NilashishC: **Bug: silently blocks "Next" with no user feedback.**

`Object.keys(tagErrors).length > 0` counts keys whose value is `''` (empty string), which `handleTagChange` stores when a user clears an input (line 190). Because `''` is falsy, neither the `TextField` error border (`!!tagErrors[index]`) nor the error `Alert` (`{tagErrors[index] && ...}`) renders — so the user has no idea why "Next" is disabled.

```suggestion
    const hasTagError = Object.values(tagErrors).some(Boolean);
```

This should be paired with fixing `handleTagChange` (see next comment) so empty entries are deleted rather than stored as `''`.

plugins/self-service/src/components/Scaffolder/ScmSelector/ScmSelectorExtension.tsx
```diff
@@ -376,8 +432,9 @@ export const ScmSelectorExtension = ({
         repoExists: false,
       });
       setRepoStatus(null);
+      resetSubmitAttempted();
```
NilashishC, line 435: **UX issue: `resetSubmitAttempted()` prematurely clears the repo-name error.**

When the user clicks "Next" with both namespace and repo empty, both tooltips appear. Selecting a namespace then fires `handleOrgChange`, which calls `resetSubmitAttempted()` — this flips `submitAttempted` back to `false` and hides the repo tooltip even though the repo is still empty. The user has to click "Next" a second time to surface that error.

The namespace tooltip naturally disappears once `selectedOrg` becomes truthy, so the reset buys nothing. Removing it lets the repo tooltip persist until the user actually fills in the field.

```suggestion
      setRepoStatus(null);
```

NilashishC (changes requested):
Good direction overall — the `FieldValidationContext` is cleanly designed, the type-toggle on `SubmitButton` is an accessible improvement over the earlier disabled-button approach, and the `ScmSelector` inline validation UX is a genuine step up. Two functional issues need addressing before this is mergeable.

**Required**

1. **`EETagsPickerExtension` silent block** (see inline on line 161): when a user clears a tag input, `handleTagChange` stores `''` at that index in `tagErrors`. Because `''` is falsy, neither the `TextField` error border nor the error `Alert` renders — but `Object.keys(tagErrors).length > 0` is `true`, so `setFieldError(fieldId, true)` fires and "Next" is silently blocked. The inline comment has a one-line suggestion fix; it should be paired with changing `handleTagChange` to delete the key rather than store `''` when `value.length === 0` (this is outside the diff so I cannot leave an inline comment there, but it is the root cause).

2. **`ScmSelectorExtension` double-click to surface repo error** (see inline on line 435): `resetSubmitAttempted()` in `handleOrgChange` causes the repo-name tooltip to vanish the moment an org is selected. The namespace tooltip disappears naturally once `selectedOrg` is truthy, so the reset is not needed.

**Recommended**

3. Hardcoded colors in `validationTooltip` / `validationIconBox` / `validationTooltipText` styles will clash on dark-themed Backstage instances. Prefer `theme.palette.background.paper`, `theme.palette.divider`, `theme.palette.warning.main`, and `theme.palette.text.primary`.

4. The two tests `'hides namespace error once org is selected'` and `'hides repo error once repository name is entered'` only assert that `onChange` was called — they do not verify the tooltip node is removed from the DOM. A regression where the tooltip persists would still pass.

NilashishC:
image

In this scenario, I would expect the user to be blocked from proceeding further. But it seems to go through with the changes in this PR.

plugins/self-service/src/components/Scaffolder/ScmSelector/ScmSelectorExtension.tsx
abhikdps: The repo name does not appear until namespace is selected, hence it does not make sense for the tooltip to appear on both namespace and repo name fields. If we do not reset it, the once the namespace is selected and repo name field appears, it would also have the tooltip which does not make sense because the user has not clicked on the Next button. The tooltip on repo name field should only appear if the field is empty and the user clicks on Next.

abhikdps:
> image
> In this scenario, I would expect the user to be blocked from proceeding further. But it seems to go through with the changes in this PR.

Fixed in [f785a0f](https://github.com/ansible/ansible-backstage-plugins/pull/277/commits/f785a0fe5620ae9ed4efe1824c296b2a89b5d907)

plugins/self-service/src/components/Scaffolder/EETagsPicker/EETagsPickerExtension.tsx
abhikdps: fixed in [ce2c848](https://github.com/ansible/ansible-backstage-plugins/pull/277/commits/ce2c848b51159d4ab5fc957d227bb4ced059e82f)

NilashishC (approved):

plugins/self-service/src/components/Scaffolder/ScmSelector/ScmSelectorExtension.tsx
NilashishC: Fair point — since the repo field is only rendered once an org is selected, removing `resetSubmitAttempted()` would cause the repo tooltip to appear immediately on field reveal, before the user has a chance to interact with it. The current behavior (require an explicit Next click to surface the repo error) is the better UX. Happy to accept the reasoning here; feel free to resolve this thread.

PR #278, opened 2026-04-24
Make source field mandatory in collections picker

abhikdps:
## Description

Make source field mandatory in collections picker. The "Add Collection" button is disabled until the source is selected after the collection name.

## Related Issues

Closes [AAP-72627](https://redhat.atlassian.net/browse/AAP-72627)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tested locally.

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **New Features**
  * Source is now required when adding or editing collections; the Add button remains disabled until both Collection and Source are selected. Added Source selection to flows that add or replace collections so items include both name and source.

* **Tests**
  * Tests updated to validate Source selection, Add-button disabled/enabled states, autocomplete/source selection flows, and payloads including source.

### DIFF
--- a/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
+++ b/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
@@ -344,11 +344,8 @@ export const CollectionsPickerExtension = ({
 
     const collectionToAdd: CollectionItem = {
       name: selectedCollection.trim(),
+      source: selectedSource,
     };
-
-    if (selectedSource) {
-      collectionToAdd.source = selectedSource;
-    }
 
     if (selectedVersion) {
       collectionToAdd.version = selectedVersion;
@@ -490,6 +487,7 @@ export const CollectionsPickerExtension = ({
                 label="Source"
                 placeholder="Select source"
                 variant="outlined"
+                required
                 className={classes.inputField}
                 InputProps={{
                   ...params.InputProps,
--- a/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
+++ b/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
@@ -338,7 +338,7 @@ export const CollectionsPickerExtension = ({
   }, [selectedCollection, selectedSource, fetchVersions]);
 
   const handleAddCollection = () => {
-    if (!selectedCollection?.trim()) {
+    if (!selectedCollection?.trim() || !selectedSource?.trim()) {
       return;
     }
 
@@ -392,7 +392,8 @@ export const CollectionsPickerExtension = ({
     setEditingIndex(index);
   };
 
-  const isAddButtonDisabled = !selectedCollection?.trim() || disabled;
+  const isAddButtonDisabled =
+    !selectedCollection?.trim() || !selectedSource?.trim() || disabled;
 
   const versionAutocompleteValue = useMemo(():
     | VersionOption
--- a/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.test.tsx
+++ b/plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.test.tsx
@@ -1730,37 +1730,6 @@ describe('CollectionsPickerExtension', () => {
   });
 
   describe('Adding Collections', () => {
-    it('adds collection with name only', async () => {
-      const mockCollections = [
-        { name: 'community.general', id: 'community.general' },
-      ];
-      mockScaffolderApi.autocomplete.mockResolvedValue({
-        results: mockCollections,
-      });
-      const props = createMockProps();
-      render(<CollectionsPickerExtension {...props} />);
-
-      await waitFor(() => {
-        expect(mockScaffolderApi.autocomplete).toHaveBeenCalled();
-      });
-
-      const collectionInput = getInputElement('Collection');
-      if (collectionInput) {
-        fireEvent.change(collectionInput, {
-          target: { value: 'community.general' },
-        });
-      }
-
-      await waitFor(() => {
-        const addButton = screen.getByRole('button', {
-          name: 'Add collection',
-        });
-        if (!addButton.hasAttribute('disabled')) {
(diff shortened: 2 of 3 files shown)


NilashishC (approved):
Clean, minimal, well-tested fix. A few small things worth addressing in a follow-up — see inline comments.

plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
```diff
@@ -338,7 +338,7 @@ export const CollectionsPickerExtension = ({
   }, [selectedCollection, selectedSource, fetchVersions]);
 
   const handleAddCollection = () => {
-    if (!selectedCollection?.trim()) {
+    if (!selectedCollection?.trim() || !selectedSource?.trim()) {
```
NilashishC, line 341: Good guard. One minor follow-up: since this now ensures `selectedSource` is non-empty before proceeding, the `if (selectedSource)` block ~8 lines below (where `collectionToAdd.source` is set) becomes dead code — that condition is always true from this point. Consider simplifying it to an unconditional assignment:

```ts
collectionToAdd.source = selectedSource;
```

plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
```diff
@@ -392,7 +392,8 @@ export const CollectionsPickerExtension = ({
     setEditingIndex(index);
   };
 
-  const isAddButtonDisabled = !selectedCollection?.trim() || disabled;
+  const isAddButtonDisabled =
+    !selectedCollection?.trim() || !selectedSource?.trim() || disabled;
```
NilashishC, line 393: Two related suggestions for follow-ups:

**1. Type definition** — `CollectionItem.source` in `types.ts` is still `source?: string` (optional). Now that the UI enforces source as mandatory, aligning the type would catch any callers constructing a `CollectionItem` without a source at compile time.

**2. Visual required indicator** — The Source `TextField` has no `required` prop. After a user selects a collection, they may not immediately understand why the Add button stays disabled. Passing `required` to the `TextField` renders MUI's standard asterisk and `aria-required`, making the expectation clear.

NilashishC (commented):
Adding inline notes — see comments on the changed lines.

plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
abhikdps: fixed in [3733839](https://github.com/ansible/ansible-backstage-plugins/pull/278/commits/37338396604a5a3f58e8f38675d148f2a5fbcdef)

plugins/self-service/src/components/Scaffolder/CollectionsPicker/CollectionsPickerExtension.tsx
abhikdps: fixed in [3733839](https://github.com/ansible/ansible-backstage-plugins/pull/278/commits/37338396604a5a3f58e8f38675d148f2a5fbcdef)

NilashishC (commented):
All three of my previous comments are addressed in this commit — thanks for the quick turnaround.

Two items from CodeRabbit's review are still open:

**1. `formData` missing required `source` (line 2226 in the test file)**

The "replaces existing collection" test seeds `formData` without a `source`:
```ts
const formData = [{ name: 'community.general', version: '1.0.0' }];
```
Now that `CollectionItem.source` is `string` (required), this fixture is inconsistent with every other updated fixture in the file. Suggested fix:
```ts
const formData: CollectionItem[] = [
  { name: 'community.general', source: 'Source 1', version: '1.0.0' },
];
```

**2. Four stale tests silently skip the add path (lines 1733, 1985, 2016, 2047)**

These tests select a collection via `fireEvent.change` but never select a source, so the Add button stays disabled. Each one has:
```ts
if (!addButton.hasAttribute('disabled')) {
  fireEvent.click(addButton); // never reached
}
```
The `click` branch is dead code — the button is always disabled at that point. Affected tests:
- `adds collection with name only` (line 1733)
- `resets form after adding collection` (line 1985)
- `handles collections array being null in handleAddCollection` (line 2016)
- `handles adding collection when collections is undefined` (line 2047)

Since `adds collection with name and source` (line 1764) and `calls onChange when adding collection with name and source` (line 2078) already cover the add path, these four can either be removed or updated to select a source before clicking Add.

abhikdps:
> All three of my previous comments are addressed in this commit — thanks for the quick turnaround.
> 
> Two items from CodeRabbit's review are still open:
> 
> **1. `formData` missing required `source` (line 2226 in the test file)**
> 
> The "replaces existing collection" test seeds `formData` without a `source`:
> 
> ```ts
> const formData = [{ name: 'community.general', version: '1.0.0' }];
> ```
> 
> Now that `CollectionItem.source` is `string` (required), this fixture is inconsistent with every other updated fixture in the file. Suggested fix:
> 
> ```ts
> const formData: CollectionItem[] = [
>   { name: 'community.general', source: 'Source 1', version: '1.0.0' },
> ];
> ```
> 
> **2. Four stale tests silently skip the add path (lines 1733, 1985, 2016, 2047)**
> 
> These tests select a collection via `fireEvent.change` but never select a source, so the Add button stays disabled. Each one has:
> 
> ```ts
> if (!addButton.hasAttribute('disabled')) {
>   fireEvent.click(addButton); // never reached
> }
> ```
> 
> The `click` branch is dead code — the button is always disabled at that point. Affected tests:
> 
> * `adds collection with name only` (line 1733)
> * `resets form after adding collection` (line 1985)
> * `handles collections array being null in handleAddCollection` (line 2016)
> * `handles adding collection when collections is undefined` (line 2047)
> 
> Since `adds collection with name and source` (line 1764) and `calls onChange when adding collection with name and source` (line 2078) already cover the add path, these four can either be removed or updated to select a source before clicking Add.

Added missing source field to formData and removed stale tests to address the review comments from coderabbitai in commit [2342c06](https://github.com/ansible/ansible-backstage-plugins/pull/278/commits/2342c06e4885a39d4d19c5c6183c6f2caa0ee662).

PR #281, opened 2026-04-27
fix: RunTask page button alignments

abhikdps:
## Description

- Extract filteredLinks before the main return so the count is available
- compute totalButtonCount (filtered links + download button + text buttons)
- Make justifyContent conditional: 'flex-start' for 1 button, 'center' for multiple
- Reuse filteredLinks in the JSX to avoid filtering twice
- Spacer <Box style={{ flex: '0 0 8.333%' }} /> — an invisible element that takes up exactly 8.33% of the container width before the button. Since flex-basis percentages resolve against the flex container's own width, this is precise.
- Result: button left edge sits at ~8.33%, button center at ~25% — exactly the midpoint between left (0%) and center (50%).
- Only when totalButtonCount === 1 - no effect on the 2+ button case.

## Related Issues

Related JIRA: [AAP-72323](https://redhat.atlassian.net/browse/AAP-72323)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tested locally

- when when totalButtonCount === 1:
Screenshot 2026-04-27 at 12 09 54

- when when totalButtonCount > 2:
Screenshot 2026-04-27 at 12 09 59

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **Refactor**
  * Improved layout and spacing in the RunTask links area for consistent alignment when showing one or multiple buttons; unified link and download button rendering for predictable behavior.
* **Tests**
  * Added tests validating button-row alignment and single-button spacing behavior for different output link scenarios.

### DIFF
--- a/plugins/self-service/src/components/RunTask/RunTask.tsx
+++ b/plugins/self-service/src/components/RunTask/RunTask.tsx
@@ -466,6 +466,24 @@ export const RunTask = () => {
     );
   }
 
+  const filteredLinks =
+    output?.links?.filter(link => {
+      if ('if' in link && link.if === false) return false;
+      if ('entityRef' in link)
+        return !!link.entityRef && link.entityRef.trim() !== '';
+      if ('url' in link) {
+        const url = link.url;
+        return !!url && url !== '#' && url.trim() !== '';
+      }
+      return false;
+    }) ?? [];
+  const textButtonCount =
+    completed && !error && output?.text && output.text.length > 0
+      ? output.text.length
+      : 0;
+  const totalButtonCount =
+    filteredLinks.length + (showDownloadButton ? 1 : 0) + textButtonCount;
+
   return (
     <Page themeId="tool">
       <Header
@@ -520,7 +538,7 @@ export const RunTask = () => {
             <Box
               display="flex"
               flexWrap="wrap"
-              justifyContent="flex-start"
+              justifyContent={totalButtonCount === 1 ? 'flex-start' : 'center'}
               style={{
                 gap: '8px',
                 flexShrink: 1,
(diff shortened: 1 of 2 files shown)


plugins/self-service/src/components/RunTask/RunTask.tsx
```diff
@@ -530,43 +548,14 @@ export const RunTask = () => {
...
-                        style={{
-                          flex: '0 0 calc(33.333% - 6px)',
-                          maxWidth: 'calc(33.333% - 6px)',
-                        }}
-                      >
-                        {link.title}
-                      </Button>
-                    );
-                  }
+              {totalButtonCount === 1 && <Box style={{ flex: '0 0 8.333%' }} />}
```
NilashishC: The `8.333%` value (= 100%/12, one column in a 12-column grid) has no inline explanation. Someone reading this code later will need to hunt down this PR description to understand the intent. A short comment here would make it self-documenting:

```suggestion
              {totalButtonCount === 1 && <Box style={{ flex: "0 0 8.333%" }} />} {/* 8.333% = 100%/12 — one grid column; shifts single button center to ~25% */}
```

plugins/self-service/src/components/RunTask/RunTask.tsx
```diff
@@ -520,7 +538,7 @@ export const RunTask = () => {
             <Box
               display="flex"
               flexWrap="wrap"
-              justifyContent="flex-start"
+              justifyContent={totalButtonCount === 1 ? 'flex-start' : 'center'}
```
NilashishC, line 542: The new conditional alignment logic (`totalButtonCount === 1 ? 'flex-start' : 'center'`) and the spacer Box on line 551 are the heart of this bug fix, but there are no tests covering either branch. Codecov reports 2 missing lines and 27 new partial branches introduced by this PR (patch coverage 91.3%).

Please add at least two test cases in `RunTask.test.tsx`:
1. **Single-button case** — mock a task with exactly one output link and assert the spacer Box is present (or query the alignment container).
2. **Multi-button case** — mock a task with 2+ output links and assert the spacer Box is absent.

These don't need to assert CSS values directly. Querying for a rendered element or using `data-testid` on the spacer Box (or the outer flex container) would be sufficient.

plugins/self-service/src/components/RunTask/RunTask.tsx
```diff
@@ -576,7 +565,23 @@ export const RunTask = () => {
...
+                    variant="contained"
+                    style={{
+                      flex: '0 0 calc(33.333% - 6px)',
+                      maxWidth: 'calc(33.333% - 6px)',
+                    }}
+                  >
+                    {link.title}
+                  </Button>
+                );
+              })}
```
NilashishC, line 590: Both branches in this `filteredLinks.map` render `<Button variant="contained" style={{ flex: "0 0 calc(33.333% - 6px)", maxWidth: "calc(33.333% - 6px)" }}>` — they differ only in `onClick` vs `href/target/rel`. Consider extracting the shared props:

```typescript
const sharedButtonProps = {
  variant: "contained" as const,
  style: { flex: "0 0 calc(33.333% - 6px)", maxWidth: "calc(33.333% - 6px)" },
};

const conditionalProps =
  "entityRef" in link && link.entityRef
    ? { onClick: () => handleEntityLinkClick(link.entityRef!) }
    : { href: link.url ?? "#", target: "_blank", rel: "noopener noreferrer" };

return <Button key={...} {...sharedButtonProps} {...conditionalProps}>{link.title}</Button>;
```

This is a low-priority nitpick — not blocking.

NilashishC (changes requested):
The fix is correct and the pre-computation of `filteredLinks` + the conditional `justifyContent` is a clean approach. One blocker before merging: the new conditional alignment logic (the core of this bug fix) has no test coverage — Codecov shows 2 missing lines and 27 new partial branches. Please add tests for the single-button and multi-button alignment cases. See inline comments for details, plus two low-priority nitpicks.

plugins/self-service/src/components/RunTask/RunTask.tsx
abhikdps: Added comment for this in [6b308ba](https://github.com/ansible/ansible-backstage-plugins/pull/281/commits/6b308bab02681e5b2746da2c4407c068b57da445)

plugins/self-service/src/components/RunTask/RunTask.tsx
abhikdps: Added tests as part of commit [6b308ba](https://github.com/ansible/ansible-backstage-plugins/pull/281/commits/6b308bab02681e5b2746da2c4407c068b57da445)

plugins/self-service/src/components/RunTask/RunTask.tsx
abhikdps: created a `sharedButtonProps` in commit : [6b308ba](https://github.com/ansible/ansible-backstage-plugins/pull/281/commits/6b308bab02681e5b2746da2c4407c068b57da445)

NilashishC (approved):
All three feedback items addressed cleanly — tests added with `data-testid` hooks, the magic number is now documented inline, and the duplicate Button props are consolidated into `sharedButtonProps`. Good to merge.

PR #305, opened 2026-04-29
Fix wrong filter type in EE catalog page

abhikdps:
## Description

Fix wrong filter type in EE catalog page

## Related Issues

Closes [AAP-73020](https://redhat.atlassian.net/browse/AAP-73020)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Tested locally.

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing tags metadata to prevent rendering errors

* **Refactor**
  * Updated execution environment filtering from tag-based to type-based selection

### DIFF
--- a/plugins/self-service/src/components/ExecutionEnvironments/catalog/CatalogContent.test.tsx
+++ b/plugins/self-service/src/components/ExecutionEnvironments/catalog/CatalogContent.test.tsx
@@ -2,7 +2,11 @@ import { render, screen, waitFor, fireEvent } from '@testing-library/react';
 import { TestApiProvider } from '@backstage/test-utils';
 import { configApiRef } from '@backstage/core-plugin-api';
 import { scmAuthApiRef } from '@backstage/integration-react';
-import { catalogApiRef } from '@backstage/plugin-catalog-react';
+import {
+  catalogApiRef,
+  EntityKindFilter,
+  EntityTypeFilter,
+} from '@backstage/plugin-catalog-react';
 import { eeBuildApiRef } from '../../../apis';
 import { NotificationProvider, notificationStore } from '../../notifications';
 import { EE_BUILD_PENDING_SESSION_KEY } from './eeBuildSession';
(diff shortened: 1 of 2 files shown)


abhikdps:
@coderabbitai review

NilashishC (approved):
The fix is correct — `EntityTypeFilter` is the right class for filtering by `spec.type`, and removing the stale `updateFilters({ tags: new EntityTagFilter(["ansible"]) })` call makes sense since the EE page does its own data fetching directly through the catalog API and the context tag filter was interfering with the star button. The null-safety additions for `entity.metadata.tags` are a nice bonus.

Two minor test quality notes in the inline comments — non-blocking.

NilashishC:
**[Minor, non-blocking] `ExecutionEnvironmentTypeFilter` test doesn't validate the specific filter type**

In `CatalogContent.test.tsx` (line 1830), the assertion only checks that `updateFilters` was called — it would pass equally with the old incorrect `EntityTagFilter`. Consider strengthening it to actually guard against regressions:

```ts
const updateFiltersCallArg = mockUpdateFilters.mock.calls[0][0];
const result = updateFiltersCallArg({});
expect(result.type).toBeInstanceOf(EntityTypeFilter);
expect(result.kind).toBeInstanceOf(EntityKindFilter);
```

---

**[Minor, non-blocking] Outdated test comment and loose assertion after the null-safety fix**

In `CatalogContent.test.tsx` (line 1465, `handles entities with missing tags property`), the comment at line 1476 still says:
> _"This will cause an error in the component, but we can test error handling"_

After the null-safety fix (`entity.metadata.tags ?? []`) added in this PR, entities with a missing `tags` property no longer cause an error — they render cleanly. The assertion on line 1494 (`expect(table !== null || error !== null).toBeTruthy()`) is also too loose now. It should be tightened to:

```ts
await waitFor(() =>
  expect(screen.getByTestId('stubbed-table-title')).toBeInTheDocument()
);
```

And the comment should be updated to reflect the corrected behaviour.

NilashishC:
Thanks for the quick follow-up @abhikdps! Both points have been addressed cleanly:

- ✅ The `ExecutionEnvironmentTypeFilter` test now imports and asserts on `EntityTypeFilter` / `EntityKindFilter` instances — this will properly catch regressions back to `EntityTagFilter`
- ✅ The outdated comment and loose assertion in the missing-tags test have been replaced with a direct, correct assertion matching the null-safety fix

PR looks good to merge. 🚀

PR #318, opened 2026-05-06
fix: restore Red Hat fonts and dark mode text after Backstage version bump

abhikdps:
## Problem

After the Backstage version bump, two visual regressions appeared across the whole app (not just our plugins):
1. **Red Hat fonts not applied** - the UI falls back to `system-ui` font-family
2.  **Dark mode text stays black** - text colour doesn't change when switching to dark mode

## Root cause

`@backstage/theme@0.7.2` silently removed automatic `<CssBaseline />` rendering from `UnifiedThemeProvider`. Previously it was rendered by default (`noCssBaseline = false`).

The RHDH theme plugin injects Red Hat font `@font-face` declarations and sets `body { font-family }` via `MuiCssBaseline.styleOverrides` inside `createComponents`. These overrides only take effect when `<CssBaseline />` is actually rendered in the tree. Without it, `@backstage/canon`'s CSS reset (`modern-normalize`) wins and sets `html { font-family: system-ui }`. Similarly, MUI's CssBaseline is responsible for setting `body { color: theme.palette.text.primary }` - in dark mode that's `#ffffff`. Without it, the body inherits the browser default (black).

### Local dev (`packages/app`)
Rendering `<CssBaseline />` from `@material-ui/core` directly in `App.tsx`, mirroring what `UnifiedThemeProvider` used to do automatically.

### RHDH dynamic plugin environment
Created an `AppThemeFixer` component (which is just `CssBaseline`) and registerd it at the `application/listener` mount point in both frontend plugins. This mount point is always active regardless of the current route, so the fix applies from the first render without requiring any host app changes.

Added to both `backstage-rhaap` and `self-service` so either plugin independently deployed in RHDH gets the fix.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed browser style inconsistencies by ensuring global style normalization applies consistently across all routes.
  * Improved theme rendering for Red Hat branding, including proper font and dark mode styling in the latest version.

### DIFF
--- a/plugins/backstage-rhaap/src/plugin.ts
+++ b/plugins/backstage-rhaap/src/plugin.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 import {
+  createComponentExtension,
   createPlugin,
   createRoutableExtension,
 } from '@backstage/core-plugin-api';
@@ -33,5 +34,15 @@ export const AnsiblePage = ansiblePlugin.provide(
     component: () =>
       import('./components/AnsiblePage').then(m => m.AnsiblePage),
     mountPoint: rootRouteRef,
+  }),
+);
+
+export const AppThemeFixer = ansiblePlugin.provide(
+  createComponentExtension({
+    name: 'AppThemeFixer',
+    component: {
+      lazy: () =>
+        import('./components/AppThemeFixer').then(m => m.AppThemeFixer),
+    },
   }),
 );
--- a/plugins/self-service/src/plugin.ts
+++ b/plugins/self-service/src/plugin.ts
@@ -48,6 +48,16 @@ export const LocationListener = selfServicePlugin.provide(
   }),
 );
 
+export const AppThemeFixer = selfServicePlugin.provide(
+  createComponentExtension({
+    name: 'AppThemeFixer',
+    component: {
+      lazy: () =>
+        import('./components/AppThemeFixer').then(m => m.AppThemeFixer),
+    },
+  }),
+);
+
 /**
  * Custom logout button that revokes the AAP OAuth token,
  * ends the AAP browser session, and signs out of Backstage.
--- a/packages/app/src/App.tsx
+++ b/packages/app/src/App.tsx
@@ -27,6 +27,7 @@ import { searchPage } from './components/search/SearchPage';
 import { Root } from './components/Root';
 import { GlobalHeader } from './components/GlobalHeader';
 import { getThemes } from '@red-hat-developer-hub/backstage-plugin-theme';
+import CssBaseline from '@material-ui/core/CssBaseline';
 
 import {
   AlertDisplay,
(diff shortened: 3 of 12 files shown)


audgirka:
@abhikdps PR looks good but I think we need to start thinking about a common `backstage-rhaap-react` front-end plugin for things like `AppThemeFixer`.

Please fix the code quality CI failure.

plugins/backstage-rhaap/src/plugin.ts
```diff
@@ -35,3 +36,13 @@ export const AnsiblePage = ansiblePlugin.provide(
     mountPoint: rootRouteRef,
   }),
 );
+
+export const AppThemeFixer = ansiblePlugin.provide(
```
NilashishC, line 40: CI is failing on **new-code coverage** (Codecov/Sonar) pointing at this new `AppThemeFixer` export.

Can we tighten the unit test to ensure this line is definitely hit under coverage (e.g., assert the export exists and `await calledWith.component.lazy()` in the test always runs), or adjust the Jest coverage collection if it’s being missed unexpectedly?

plugins/self-service/src/plugin.ts
```diff
@@ -48,6 +48,16 @@ export const LocationListener = selfServicePlugin.provide(
   }),
 );
 
+export const AppThemeFixer = selfServicePlugin.provide(
```
NilashishC, line 51: Same as the other plugin: Codecov/Sonar flags **new-code coverage** missing around the new `AppThemeFixer` component extension.

Please tweak/add a test assertion so this export path is unambiguously exercised under coverage (or align coverage config if tests already execute it).

NilashishC (approved):
Re-review: ✅ SonarCloud and test checks are now green.

The `CssBaseline` fix and use of `application/listener` mountpoint look correct for restoring fonts + dark-mode text after the Backstage theme bump.

(I left two inline notes near the `AppThemeFixer` exports based on the earlier coverage failure; feel free to ignore them now if the coverage gating is fully resolved.)

plugins/backstage-rhaap/src/plugin.ts
abhikdps: done in [89df87f](https://github.com/ansible/ansible-backstage-plugins/pull/318/commits/89df87f2fb335636a42e431466c571d9d8c15c63)

plugins/self-service/src/plugin.ts
abhikdps: done in [89df87f](https://github.com/ansible/ansible-backstage-plugins/pull/318/commits/89df87f2fb335636a42e431466c571d9d8c15c63)

PR #356, opened 2026-05-18
feat: add paginated caching for EE and Git Repos pages + update CI activity endpoint

abhikdps:
## Description

- Paginated entity caching for EE and Git Repos pages - both pages now use the same cursor-based queryEntities pagination pattern as Collections, fetching an initial 50 items then loading remaining entities in background batches of 200. Previously both pages fetched all entities in a single unbounded getEntities call.
- Extracted shared cache infrastructure — created a generic PaginatedEntityCache<TState> base class and shared hooks (useCacheSubscription, usePagination) under common/cache/, eliminating ~1,000 lines of  duplicated loading, subscription, epoch-guard, TTL, and pagination logic across the Collections, EE, and Git Repos caches.
- Updated CI activity endpoint - updated the per-entity GET /ansible/git/ci-activity endpoint to make batch requests. Both useLatestCIActivity (summary row) and RepositoriesCIActivityTab (full activity table) now send one request for all repos instead of N+1.

## Test plan

- Verify EE page loads with paginated data, filters by owner/tag, and paginates correctly
- Verify Git Repos page loads with paginated data, filters by source, and paginates correctly
- Verify Git Repos page CI activity column shows latest activity without repeated API calls (check network tab — should be a single POST to /ansible/git/ci-activity)
- Verify CI Activity tab loads and displays workflow runs / pipelines from mixed GitHub + GitLab repos
- Verify Collections page still works (cache refactored to use shared base class)
- Verify sync polling invalidates git repos cache (trigger a PAH sync and confirm git repos page refreshes)
- yarn tsc, yarn lint, yarn test pass

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

WIP

## Checklist

- [ ] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **New Features**
  * Batched CI activity lookup: fetch CI status for many repositories in a single request with per-item success/error results.
  * Shared paginated caches and hooks powering Collections, Execution Environments, and Git Repositories with consistent filtering and client-side pagination.

* **Performance**
  * TTL-backed background caching with epoch-based invalidation reduces repeated catalog requests and speeds UI loads.
  * Cache invalidation wired into sync flows for fresher data.

* **Tests**
  * Updated tests to cover batch CI protocol, cache behavior, pagination, and retry/cancellation scenarios.

### DIFF
--- a/plugins/catalog-backend-module-rhaap/src/router.ts
+++ b/plugins/catalog-backend-module-rhaap/src/router.ts
@@ -51,8 +51,8 @@ import {
   createRequireSuperuserMiddleware,
   createRequireUserOrExternalAccessMiddleware,
   createPermissionCheckMiddleware,
-  handleGitHubCIActivity,
-  handleGitLabCIActivity,
+  fetchGitHubCIActivityData,
+  fetchGitLabCIActivityData,
   parseEeBuildRequestBody,
   validateGitHubHost,
   isKnownEeBuildError,
@@ -799,59 +799,133 @@ export async function createRouter(options: {
     }
   });
 
-  // Unified CI activity proxy for GitHub and GitLab
-  // Query params:
+  // Batch CI activity proxy for GitHub Actions and GitLab Pipelines.
+  // POST body: { items: [{ key, provider, owner?, repo?, host?, projectPath?, per_page? }] }
   //   - provider: 'github' | 'gitlab' (required)
-  //   - host: hostname (optional, defaults based on provider)
-  //   - per_page: number of results (optional)
+  //   - host: SCM hostname (optional, defaults to github.com / gitlab.com)
+  //   - per_page: max results per repo (optional, default 15, max 100)
   //   GitHub-specific: owner, repo
   //   GitLab-specific: projectPath
-  router.get('/ansible/git/ci-activity', async (request, response) => {
-    const credentials = await httpAuth.credentials(request as any);
+  // Response: { results: { [key]: { status, data } | { error } } }
+  router.post(
+    '/ansible/git/ci-activity',
+    express.json(),
+    async (request, response) => {
+      const credentials = await httpAuth.credentials(request as any);
 
-    const [[gitRepoDecision], [catalogReadDecision]] = await Promise.all([
-      permissions.authorize([{ permission: gitRepositoriesViewPermission }], {
-        credentials,
-      }),
-      permissions.authorizeConditional(
-        [{ permission: catalogEntityReadPermission }],
-        { credentials },
-      ),
-    ]);
+      const [[gitRepoDecision], [catalogReadDecision]] = await Promise.all([
+        permissions.authorize(
+          [{ permission: gitRepositoriesViewPermission }],
+          { credentials },
+        ),
+        permissions.authorizeConditional(
+          [{ permission: catalogEntityReadPermission }],
+          { credentials },
+        ),
+      ]);
 
-    const hasGitRepoView = gitRepoDecision.result === AuthorizeResult.ALLOW;
-    const hasCatalogRead = catalogReadDecision.result !== AuthorizeResult.DENY;
+      const hasGitRepoView = gitRepoDecision.result === AuthorizeResult.ALLOW;
+      const hasCatalogRead =
+        catalogReadDecision.result !== AuthorizeResult.DENY;
 
-    if (!hasGitRepoView || !hasCatalogRead) {
-      response
-        .status(403)
-        .json({ error: 'Forbidden: insufficient permissions' });
-      return;
-    }
+      if (!hasGitRepoView || !hasCatalogRead) {
+        response
+          .status(403)
+          .json({ error: 'Forbidden: insufficient permissions' });
+        return;
+      }
 
-    const provider = (request.query.provider as string)?.toLowerCase();
-    const perPage = Math.min(Number(request.query.per_page) || 15, 100);
-    const ciActivityDeps = {
-      config,
-      logger,
-      scmIntegrations: scmClientFactory.integrations,
-      githubCredentialsProvider: scmClientFactory.githubCredentialsProvider,
-    };
+      const { items } = request.body as {
+        items?: Array<{
+          key: string;
+          provider: string;
+          owner?: string;
+          repo?: string;
+          host?: string;
+          projectPath?: string;
+          per_page?: number;
+        }>;
+      };
 
-    if (!provider || !['github', 'gitlab'].includes(provider)) {
-      response.status(400).json({
-        error:
-          "Missing or invalid 'provider' query parameter. Must be 'github' or 'gitlab'.",
-      });
-      return;
-    }
+      if (!Array.isArray(items) || items.length === 0) {
+        response
+          .status(400)
+          .json({ error: "'items' must be a non-empty array" });
+        return;
+      }
 
-    if (provider === 'github') {
-      await handleGitHubCIActivity(ciActivityDeps, request, response, perPage);
-    } else {
-      await handleGitLabCIActivity(ciActivityDeps, request, response, perPage);
-    }
-  });
+      if (items.length > 100) {
+        response
+          .status(400)
+          .json({ error: 'Maximum 100 items per batch request' });
+        return;
+      }
+
+      const ciActivityDeps = {
+        config,
+        logger,
+        scmIntegrations: scmClientFactory.integrations,
+        githubCredentialsProvider: scmClientFactory.githubCredentialsProvider,
+      };
+
+      const CONCURRENCY_LIMIT = 5;
+      const results: Record<
+        string,
+        { status: number; data: any } | { error: string }
+      > = {};
+
+      // Process items with concurrency limit
+      let index = 0;
+      const processNext = async (): Promise<void> => {
+        while (index < items.length) {
+          const currentIndex = index++;
+          const item = items[currentIndex];
+          const prov = item.provider?.toLowerCase();
+          const perPage = Math.min(Number(item.per_page) || 15, 100);
+
+          try {
+            if (prov === 'github') {
+              const result = await fetchGitHubCIActivityData(ciActivityDeps, {
+                owner: item.owner ?? '',
+                repo: item.repo ?? '',
+                host: item.host ?? 'github.com',
+                perPage,
+              });
+              results[item.key] =
+                'error' in result
+                  ? { error: result.error }
+                  : { status: result.status, data: result.data };
+            } else if (prov === 'gitlab') {
+              const result = await fetchGitLabCIActivityData(ciActivityDeps, {
+                projectPath: item.projectPath ?? '',
+                host: item.host ?? 'gitlab.com',
+                perPage,
+              });
+              results[item.key] =
+                'error' in result
+                  ? { error: result.error }
+                  : { status: result.status, data: result.data };
+            } else {
+              results[item.key] = { error: `Unknown provider: ${prov}` };
+            }
+          } catch (err) {
+            results[item.key] = {
+              error:
+                err instanceof Error ? err.message : 'Unknown error occurred',
+            };
+          }
+        }
+      };
+
+      const workers = Array.from(
+        { length: Math.min(CONCURRENCY_LIMIT, items.length) },
+        () => processNext(),
+      );
+      await Promise.all(workers);
+
+      response.status(200).json({ results });
+    },
+  );
 
   return router;
 }
--- a/plugins/catalog-backend-module-rhaap/src/router.ts
+++ b/plugins/catalog-backend-module-rhaap/src/router.ts
@@ -875,6 +875,8 @@ export async function createRouter(options: {
         seenKeys.add(item.key);
       }
 
+      logger.info(`[ci-activity] Batch request: ${items.length} items`);
+
       const ciActivityDeps = {
         config,
         logger,
@@ -885,12 +887,12 @@ export async function createRouter(options: {
       const CONCURRENCY_LIMIT = 5;
       const results: Record<
         string,
-        { status: number; data: any } | { error: string }
+        { status: number; data: unknown } | { error: string }
       > = {};
 
       const processItem = async (
         item: (typeof items)[number],
-      ): Promise<{ status: number; data: any } | { error: string }> => {
+      ): Promise<{ status: number; data: unknown } | { error: string }> => {
         const prov = item.provider?.toLowerCase();
         const perPage = Math.min(Number(item.per_page) || 15, 100);
 
@@ -921,15 +923,16 @@ export async function createRouter(options: {
       let index = 0;
       const processNext = async (): Promise<void> => {
         while (index < items.length) {
+          // safe: single-threaded JS, ++ completes before await
           const currentIndex = index++;
           const item = items[currentIndex];
           try {
             results[item.key] = await processItem(item);
           } catch (err) {
-            results[item.key] = {
-              error:
-                err instanceof Error ? err.message : 'Unknown error occurred',
-            };
+            const msg =
+              err instanceof Error ? err.message : 'Unknown error occurred';
+            logger.warn(`[ci-activity] Item '${item.key}' failed: ${msg}`);
+            results[item.key] = { error: msg };
           }
         }
       };
@@ -939,6 +942,13 @@ export async function createRouter(options: {
         () => processNext(),
       );
       await Promise.all(workers);
+
+      const errorCount = Object.values(results).filter(
+        r => 'error' in r,
+      ).length;
+      if (errorCount > 0) {
+        logger.warn(`[ci-activity] ${errorCount}/${items.length} items failed`);
+      }
 
       response.status(200).json({ results });
     },
--- a/plugins/catalog-backend-module-rhaap/src/router.ts
+++ b/plugins/catalog-backend-module-rhaap/src/router.ts
@@ -860,6 +860,21 @@ export async function createRouter(options: {
         return;
       }
 
+      const seenKeys = new Set<string>();
+      for (const item of items) {
+        if (!item.key || typeof item.key !== 'string') {
+          response
+            .status(400)
+            .json({ error: "Each item must have a non-empty 'key'" });
+          return;
+        }
+        if (seenKeys.has(item.key)) {
+          response.status(400).json({ error: `Duplicate key: '${item.key}'` });
+          return;
+        }
+        seenKeys.add(item.key);
+      }
+
       const ciActivityDeps = {
         config,
         logger,
--- a/plugins/catalog-backend-module-rhaap/src/router.ts
+++ b/plugins/catalog-backend-module-rhaap/src/router.ts
@@ -814,10 +814,9 @@ export async function createRouter(options: {
       const credentials = await httpAuth.credentials(request as any);
 
       const [[gitRepoDecision], [catalogReadDecision]] = await Promise.all([
-        permissions.authorize(
-          [{ permission: gitRepositoriesViewPermission }],
-          { credentials },
-        ),
+        permissions.authorize([{ permission: gitRepositoriesViewPermission }], {
+          credentials,
+        }),
         permissions.authorizeConditional(
           [{ permission: catalogEntityReadPermission }],
           { credentials },
--- a/plugins/catalog-backend-module-rhaap/src/router.ts
+++ b/plugins/catalog-backend-module-rhaap/src/router.ts
@@ -873,40 +873,43 @@ export async function createRouter(options: {
         { status: number; data: any } | { error: string }
       > = {};
 
-      // Process items with concurrency limit
+      const processItem = async (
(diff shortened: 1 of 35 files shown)


abhikdps:
@coderabbitai review

abhikdps:
@coderabbitai review

plugins/self-service/src/components/common/cache/useCacheSubscription.ts
```diff
@@ -0,0 +1,125 @@
...
+          setError(state.error);
+          setInitialLoading(false);
+        }
+        onCacheUpdateRef.current?.(state);
+      }
+    });
+
+    return unsubscribe;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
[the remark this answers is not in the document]
abhikdps: I don't think this is applicable. Cache is a module-level singleton constant and hence, it never changes identity. Adding cache to deps would be dead code.

abhikdps:
@coderabbitai review

plugins/self-service/src/components/common/cache/useCacheSubscription.ts
```diff
@@ -0,0 +1,128 @@
...
+    const unsubscribe = cache.subscribe(state => {
+      if (isMountedRef.current) {
+        setAllEntities(state.entities);
+        setCacheState(state);
+        setLoadingMore(!state.isFullyLoaded && cache.isLoading());
+        if (state.error) {
+          setError(state.error);
+          setInitialLoading(false);
+        }
+        onCacheUpdateRef.current?.(state);
```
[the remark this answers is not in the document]
abhikdps: fixed in [8be8cde](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/8be8cdef79334a7adae8d358ba18677b2574c502). check if the fix is correct and resolve conversation it is.

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
```diff
@@ -0,0 +1,285 @@
...
+    this.loadingPromise = this.performInitialLoad();
+    await this.loadingPromise;
+  }
+
+  private async performInitialLoad(): Promise<void> {
+    if (!this.catalogApi) return;
+    const epoch = this.fetchEpoch;
+
+    try {
+      const response = await this.catalogApi.queryEntities({
```
ganeshrn, line 111: **Unstable offset pagination without sort order.** `queryEntities` with `offset` but no `orderFields` can return inconsistent results if entities are added/removed between pages (duplicates or missed items). At the scale this fix targets (10,000+ entities), this is a real risk during concurrent catalog ingestion.

Same issue at line 196.

Suggested fix:
```typescript
const response = await this.catalogApi.queryEntities({
  filter: this.config.entityFilter,
  limit: INITIAL_FETCH_LIMIT,
  orderFields: [{ field: 'metadata.uid', order: 'asc' }],
});
```
`metadata.uid` is unique per entity, so sort order is stable across pages.

ganeshrn (commented):
Review for AAP-75647 (critical, release-blocker). The fix correctly addresses the root cause: unbounded `getEntities` calls saturating the Node.js event loop under concurrent load. Paginated caching + batch CI endpoint is the right approach. Nine informational findings, no blockers. Suggested inline fixes included.

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
```diff
@@ -0,0 +1,285 @@
...
+          limit: BACKGROUND_FETCH_LIMIT,
+          offset,
+        });
+
+        if (epoch !== this.fetchEpoch) break;
+
+        const newItems = response?.items || [];
+        if (newItems.length === 0) break;
+
+        allItems = [...allItems, ...newItems];
```
ganeshrn: **O(n²) array copies in background fetch loop.** Each iteration does `allItems = [...allItems, ...newItems]`, creating a full copy. With 10,000 entities in 200-item batches, that's ~50 iterations of progressively larger copies.

Suggested fix:
```typescript
allItems.push(...newItems); // mutate in place, no copy
```
The state update on line 210 already creates a new object via spread, so immutability is preserved where React needs it.

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -799,59 +799,150 @@ export async function createRouter(options: {
...
+            ? { error: result.error }
+            : { status: result.status, data: result.data };
+        }
+        return { error: `Unknown provider: ${prov}` };
+      };
+
+      let index = 0;
+      const processNext = async (): Promise<void> => {
+        while (index < items.length) {
+          const currentIndex = index++;
```
ganeshrn, line 927: **Safe but non-obvious concurrent index increment.** `const currentIndex = index++` inside async workers looks like a race condition. It's actually safe in single-threaded Node.js (the `++` completes synchronously before any `await`), but future readers may question it.

Suggested: add a one-line comment:
```typescript
const currentIndex = index++; // safe: single-threaded JS, ++ completes before await
```

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -799,59 +799,150 @@ export async function createRouter(options: {
...
+        config,
+        logger,
+        scmIntegrations: scmClientFactory.integrations,
+        githubCredentialsProvider: scmClientFactory.githubCredentialsProvider,
+      };
+
+      const CONCURRENCY_LIMIT = 5;
+      const results: Record<
+        string,
+        { status: number; data: any } | { error: string }
```
ganeshrn: **`data: any` weakens type safety on API response contract.** Since this is a public API response, consider `unknown` to prevent accidental untyped property access.

Suggested fix:
```typescript
const results: Record<
  string,
  { status: number; data: unknown } | { error: string }
> = {};
```
Same on line 893.

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
```diff
@@ -0,0 +1,285 @@
+import { Entity } from '@backstage/catalog-model';
```
ganeshrn, line 1: **No unit tests for the shared cache base class.** `PaginatedEntityCache` is the foundation all three page caches depend on, but it has no test file. The concrete caches (collections, EE, git repos) have tests, but the base class logic — epoch invalidation, TTL expiry, background loading recovery, error handling during background fetch, concurrent `startLoading` calls — is not directly tested.

Consider adding `PaginatedEntityCache.test.ts` covering:
- Initial load populates state and notifies listeners
- Background loading appends pages correctly
- Epoch invalidation aborts stale fetches
- TTL expiry returns null from `getState()`
- Error during background fetch preserves partial data
- Concurrent `startLoading` calls don't duplicate fetches

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
```diff
@@ -54,6 +55,102 @@ async function fetchWithRetry(
...
+): Record<string, LatestActivityEntry> {
+  const map: Record<string, LatestActivityEntry> = {};
+
+  for (const [key, result] of Object.entries(results)) {
+    if ('error' in result) {
+      map[key] = { text: NO_ACTIVITY };
+      continue;
+    }
+
+    const item = batchItems.find(i => i.key === key);
```
ganeshrn: **O(n) `.find()` inside a loop — quadratic at batch max.** With 100 items (the batch limit), this does up to 10,000 iterations.

Suggested fix:
```typescript
function buildActivityMap(
  batchItems: BatchItem[],
  results: Record<string, { status: number; data: any } | { error: string }>,
): Record<string, LatestActivityEntry> {
  const map: Record<string, LatestActivityEntry> = {};
  const itemsByKey = new Map(batchItems.map(i => [i.key, i]));

  for (const [key, result] of Object.entries(results)) {
    if ('error' in result) {
      map[key] = { text: NO_ACTIVITY };
      continue;
    }
    const item = itemsByKey.get(key);
    // ... rest stays the same
```

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -799,59 +799,150 @@ export async function createRouter(options: {
...
-  //   - per_page: number of results (optional)
+  //   - host: SCM hostname (optional, defaults to github.com / gitlab.com)
+  //   - per_page: max results per repo (optional, default 15, max 100)
   //   GitHub-specific: owner, repo
   //   GitLab-specific: projectPath
-  router.get('/ansible/git/ci-activity', async (request, response) => {
-    const credentials = await httpAuth.credentials(request as any);
+  // Response: { results: { [key]: { status, data } | { error } } }
+  router.post(
+    '/ansible/git/ci-activity',
```
ganeshrn, line 811: **GET→POST breaking API change on `/ansible/git/ci-activity`.** The OpenAPI spec was updated (good), but any existing consumers of the GET endpoint will break silently. Worth confirming no external consumers. If external consumers exist for example ATF test, consider updating the test cases.

plugins/self-service/src/components/notifications/syncPollingService.ts
```diff
@@ -113,6 +114,11 @@ class SyncPollingService {
     return gen === this.pollGeneration;
   }
 
+  private invalidateAllCaches(): void {
```
ganeshrn, line 118: **`eeCache` not invalidated on sync completion.** `invalidateAllCaches()` invalidates collections and git repos caches, but not `eeCache`. If EE entities can be affected by sync operations (the `/ansible/ee` POST route suggests EEs can be registered programmatically), the EE cache would show stale data until its 5-minute TTL expires.

Suggested fix:
```typescript
import { eeCache } from '../ExecutionEnvironments/catalog/eeCache';

private invalidateAllCaches(): void {
  collectionsCache.invalidateFetchedData();
  gitReposCache.invalidateFetchedData();
  eeCache.invalidateFetchedData();
}
```

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
```diff
@@ -0,0 +1,285 @@
...
+import { CatalogApi } from '@backstage/plugin-catalog-react';
+import { BaseCacheState, CacheUpdateListener, CacheConfig } from './types';
+
+const CACHE_TTL_MS = 5 * 60 * 1000;
```
ganeshrn, line 5: **Jira AC cross-reference (AAP-75647):**

- **AC #1** (10+ concurrent, no TCP drops): ✅ Addressed — paginated 50+200 batches dramatically reduce per-request event loop blocking vs unbounded fetch. Needs load test verification.
- **AC #2** (proper HTTP responses, not silent disconnect): ⚠️ Not directly addressed at plugin level — this is a Node.js/Backstage infrastructure concern. The fix reduces the load that causes saturation, which is the right approach, but no explicit backpressure/503 handling was added.
- **AC #3** (429/503 vs connection close): ⚠️ Same as AC #2.
- **AC #4** (< 3s P95 at 4 users): ✅ Initial load of 50 entities should be well under 3s vs the previous unbounded fetch (7-70s under load).

Consider noting in the PR description that AC #2/#3 are addressed by load reduction rather than explicit overload handling.

plugins/catalog-backend-module-rhaap/src/router.ts
ganeshrn, line 811: **Batch CI activity handler has zero logging.** Every other POST endpoint in this file logs request entry, key actions, and errors. This handler is silent — makes production debugging harder when CI activity requests fail or slow down.

Suggested additions:
```typescript
// After auth checks and validation, before processing:
logger.info(`[ci-activity] Batch request: ${items.length} items`);

// In the catch block inside processNext (line 928):
} catch (err) {
  const msg = err instanceof Error ? err.message : 'Unknown error occurred';
  logger.warn(`[ci-activity] Item '${item.key}' failed: ${msg}`);
  results[item.key] = { error: msg };
}

// Before response.status(200):
const errorCount = Object.values(results).filter(r => 'error' in r).length;
if (errorCount > 0) {
  logger.warn(`[ci-activity] ${errorCount}/${items.length} items failed`);
}
```
Three lines. Entry, per-item errors, summary. Matches the pattern used by the sync endpoints in the same file.

ganeshrn (commented):
Follow-up: logging gap in the batch CI activity handler.

plugins/catalog-backend-module-rhaap/src/router.ts
abhikdps: added the suggested change in [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/self-service/src/components/notifications/syncPollingService.ts
abhikdps: fixed in [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
abhikdps: Added `orderFields: [{field: 'metadata.uid', order: 'asc'}]` to both queryEntities calls as part of commit [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/catalog-backend-module-rhaap/src/router.ts
abhikdps: added comment as part of commit [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
abhikdps: Created test suite for PaginatedEntityCache base class. Please check the new file in the changes.

plugins/self-service/src/components/common/cache/PaginatedEntityCache.ts
abhikdps: Changed `allItems = [...allItems, ...newItems]` to `allItems.push(...newItems)` as part of commit [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/catalog-backend-module-rhaap/src/router.ts
abhikdps: Changed `data: any` to `data: unknown` in results types as part of commit [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
abhikdps: Replaced `.find()` loop with map lookup in `buildActivityMap` as part of [a4ab8d9](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/a4ab8d9d986def377f046c72e919ba773eca089e)

plugins/self-service/src/components/RunTask/RunTask.test.tsx
```diff
@@ -1979,24 +1984,31 @@ describe('RunTask', () => {
 
       mockCatalogApi.getEntityByRef.mockResolvedValue(null);
 
-      await render(<RunTask />);
+      await act(async () => {
+        render(<RunTask />);
+      });
```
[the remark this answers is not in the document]
abhikdps: `render()` in React Testing Library is synchronous, not async. adding await is harmless but unnecessary.

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
```diff
@@ -54,6 +56,161 @@ async function fetchWithRetry(
   return null;
 }
 
+interface BatchItem {
```
ganeshrn: **`BatchItem` interface and `buildBatchItems` are duplicated.** This same interface exists in `RepositoriesCIActivityTab.tsx:82` and the build function is nearly identical (`buildBatchItems` here vs `buildCIBatchItems` there, same entity → GitHub/GitLab mapping logic).

Suggested: extract to a shared `ciBatchUtils.ts` in this directory:
- `BatchItem` interface (with `per_page` optional)
- `buildBatchItems(entities, perPage = 1)` — unified, covers both use cases

~60 lines of duplication removed, no behavior change.

plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
```diff
@@ -78,202 +79,239 @@ const STATUS_LABELS: Record<CIActivityRow['status'], string> = {
...
+  ];
+  const s = raw.toLowerCase();
+  return valid.includes(s as CIActivityRow['status'])
+    ? (s as CIActivityRow['status'])
+    : 'unknown';
+}
+
+function parseGitHubRuns(
+  key: string,
+  data: any,
```
ganeshrn: **`data: any` in 5 parse functions.** Lines 149, 158, 173, 182, 196 use `any` for CI activity response data. The router was already fixed to `data: unknown`. Frontend consumers should match.

Suggested: change to `unknown` and add type guards:
```typescript
function parseGitHubRuns(
  key: string,
  data: unknown,
  entity: Entity,
): CIActivityRow[] {
  const runs = Array.isArray((data as Record<string, unknown>)?.workflow_runs)
    ? (data as Record<string, unknown>).workflow_runs as Record<string, unknown>[]
    : [];
```
Same treatment for `parseGitLabPipelines`. Also applies to `parseGitHubActivity` and `parseGitLabActivity` in `useLatestCIActivity.ts:98,109`.

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
```diff
@@ -1,21 +1,22 @@
...
 
 const NO_ACTIVITY = 'N/A';
-const GITHUB_BATCH_SIZE = 3;
-const GITHUB_BATCH_DELAY_MS = 100;
-const GITLAB_BATCH_SIZE = 2;
-const GITLAB_BATCH_DELAY_MS = 150;
 const MAX_RETRIES = 2;
 const INITIAL_RETRY_DELAY_MS = 300;
 const RETRYABLE_STATUS_CODES = new Set([502, 503, 504]);
+const BATCH_CHUNK_SIZE = 100;
```
ganeshrn: **`BATCH_CHUNK_SIZE` and `PARALLEL_LIMIT` are hardcoded in two files with the same values.** `BATCH_CHUNK_SIZE = 100` here, `CHUNK_SIZE = 100` at `RepositoriesCIActivityTab.tsx:271`. `PARALLEL_LIMIT = 5` in both files (line 172 here, line 272 there).

Suggested: move both to `GitRepositories/constants.ts` alongside the existing `PAGE_SIZE`:
```typescript
export const CI_BATCH_CHUNK_SIZE = 100;
export const CI_PARALLEL_LIMIT = 5;
```
One place to tune, no behavior change.

ganeshrn (commented):
Follow-up: three code quality improvements — no functionality changes, pure refactoring.

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
abhikdps: moved `BatchItem` interface and `buildBatchItems` to a new shared `ciBatchUtils.ts` file in commit: [85602c3](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/85602c333bfd7cd795356d28b8545cc1c706a9c8)

plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
abhikdps: Added type guards and changed data type to unknown everywhere mentioned as part of commit: [85602c3](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/85602c333bfd7cd795356d28b8545cc1c706a9c8)

plugins/self-service/src/components/GitRepositories/useLatestCIActivity.ts
abhikdps: Done as part of [85602c3](https://github.com/ansible/ansible-backstage-plugins/pull/356/commits/85602c333bfd7cd795356d28b8545cc1c706a9c8)

ganeshrn (approved):
All 13 review findings addressed across 3 rounds. Re-reviewed the full PR in its current state:

LGTM.

PR #359, opened 2026-05-20
[AAP-77602] feat: cache CI activity data across tab switches

abhikdps:
## Description

The CI Activity tab re-fetches all data every time the user navigates away and back because conditional rendering unmounts the component, destroying in-flight requests and local state. Introduce a singleton ciActivityCache that persists across mount/unmount cycles with a 3-min TTL, streams incremental results to subscribers, and deduplicates concurrent fetches.

- Extract pure parsing functions to ciActivityUtils.ts
- Add ciActivityCache.ts singleton with subscribe/notify pattern
- Simplify RepositoriesCIActivityTab to consume cache state
- Clear cache between tests to prevent state leakage

## Summary

- CI Activity tab data now survives tab switches - returning within 3 minutes shows cached results instantly instead of re-fetching all chunks
- Background fetches continue even when the user navigates away, so partial results are available on return
- Extracted CI parsing logic (status normalization, row building) into `ciActivityUtils.ts` for reuse between the cache and component

## Related Issues

Related JIRA: [AAP-77602](https://redhat.atlassian.net/browse/AAP-77602)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **Refactor**
  * Centralized CI activity fetching, parsing, caching and subscription-based updates for consistent UI state and improved performance; repository CI tab now consumes this shared cache and utilities.
  * Normalized CI status labels and unified row-building with sorting and a 150-row limit.

* **Tests**
  * Added unit tests for cache behavior (TTL expiry, invalidate, subscriptions), parsing utilities, batch failure handling, and clearing cache between tests to avoid cross-test leakage.

### DIFF
--- a/plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
+++ b/plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
@@ -55,6 +55,7 @@ export const RepositoriesCIActivityTab = ({
   const [triggerFilter, setTriggerFilter] = useState<string>('All');
 
   useEffect(() => {
+    let cancelled = false;
     const unsubscribe = ciActivityCache.subscribe(setCacheState);
 
     if (filterByEntity) {
@@ -67,19 +68,24 @@ export const RepositoriesCIActivityTab = ({
           filter: [{ kind: 'Component', 'spec.type': 'git-repository' }],
         })
         .then(response => {
+          if (cancelled) return;
           const items = Array.isArray(response)
             ? response
             : (response?.items ?? []);
           ciActivityCache.startLoading(items, discoveryApi, fetchApi);
         })
         .catch(e => {
+          if (cancelled) return;
           ciActivityCache.setError(
             e instanceof Error ? e.message : 'Failed to load entities',
           );
         });
     }
 
-    return unsubscribe;
+    return () => {
+      cancelled = true;
+      unsubscribe();
+    };
   }, [discoveryApi, fetchApi, catalogApi, filterByEntity, cachedEntities]);
 
   const rows = useMemo(() => cacheState?.rows ?? [], [cacheState?.rows]);
--- a/plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
+++ b/plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Box, Link, Paper, Typography } from '@material-ui/core';
 import { useTheme } from '@material-ui/core/styles';
 import Autocomplete from '@material-ui/lab/Autocomplete';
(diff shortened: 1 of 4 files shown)


plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
```diff
@@ -200,88 +48,44 @@ export const RepositoriesCIActivityTab = ({
...
-      } else {
-        const response = await catalogApi.getEntities({
+  useEffect(() => {
+    const unsubscribe = ciActivityCache.subscribe(setCacheState);
+
+    if (filterByEntity) {
+      ciActivityCache.startLoading([filterByEntity], discoveryApi, fetchApi);
+    } else if (cachedEntities && cachedEntities.length > 0) {
+      ciActivityCache.startLoading(cachedEntities, discoveryApi, fetchApi);
+    } else {
```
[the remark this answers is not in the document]
abhikdps: Not valid. empty array never passed in practice

plugins/self-service/src/components/GitRepositories/RepositoriesCIActivityTab.tsx
```diff
@@ -200,88 +48,44 @@ export const RepositoriesCIActivityTab = ({
...
-    try {
-      let entities: Entity[];
-      if (filterByEntity) {
-        entities = [filterByEntity];
-      } else if (cachedEntities && cachedEntities.length > 0) {
-        entities = cachedEntities;
-      } else {
-        const response = await catalogApi.getEntities({
+  useEffect(() => {
+    const unsubscribe = ciActivityCache.subscribe(setCacheState);
```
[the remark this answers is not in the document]
abhikdps: Not a real issue - global page and detail page are separate routes.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,190 @@
...
+        timestamp: Date.now(),
+      };
+      notify();
+    } finally {
+      fetchPromise = null;
+    }
+  })();
+
+  fetchPromise = promise;
+}
```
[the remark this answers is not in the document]
abhikdps: Not valid. fire-and-forget by design, subscribers get results via notify.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,205 @@
...
+  rows: CIActivityRow[];
+  loading: boolean;
+  fetchingMore: boolean;
+  error: string | null;
+  timestamp: number;
+};
+
+type Listener = (state: CIActivityCacheState) => void;
+
+let state: CIActivityCacheState | null = null;
```
ganeshrn, line 19: **Cache has no entity identity — serves wrong data when entity set changes.** `CIActivityCacheState` stores `rows`, `loading`, `error`, `timestamp` but not *which entities* the rows came from. Two scenarios break:

1. User views GitHub provider detail (CI tab loads with `filterByEntity=github-repo-1`). Cache stores 1 repo's CI data. User navigates to the full CI Activity page (`cachedEntities=all 50 repos`). Cache TTL hasn't expired. `getState()` returns single-repo data as if it were the full dataset.

2. User navigates from one provider detail to another while a fetch is in-flight. `startLoading` returns early (line 66: `if (fetchPromise !== null) return;`), so the second entity set is silently dropped.

Suggested fix — add an entity key and compare on load:
```typescript
// Add to CIActivityCacheState:
export type CIActivityCacheState = {
  rows: CIActivityRow[];
  loading: boolean;
  fetchingMore: boolean;
  error: string | null;
  timestamp: number;
  entityKey: string; // identity of the entity set this data belongs to
};

// At the top of startLoading, compute and compare:
async function startLoading(
  entities: Entity[],
  discoveryApi: DiscoveryApi,
  fetchApi: FetchApi,
): Promise<void> {
  const entityKey = entities
    .map(e => stringifyEntityRef(e))
    .sort()
    .join('\0');

  const cached = getState();
  // Same entities, valid cache — skip
  if (cached && cached.entityKey === entityKey && !cached.loading && !cached.fetchingMore) {
    return;
  }

  // Different entities or no cache — invalidate stale data and restart
  if (cached?.entityKey !== entityKey || fetchPromise !== null) {
    generation += 1;
    state = null;
    fetchPromise = null;
  }

  // ... rest of startLoading continues with the new entity set
```
This ensures the cache only returns data that matches the current entity set, and aborts stale fetches when the entity set changes.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  const cached = getState();
+  if (cached && !cached.loading && !cached.fetchingMore) return;
+
+  const { items: batchItems, entityMap } = buildBatchItems(entities, 15);
+
+  if (batchItems.length === 0) {
+    state = {
+      rows: [],
+      loading: false,
```
ganeshrn, line 81: **Inflight guard drops new entity sets instead of invalidating.** `if (fetchPromise !== null) return;` silently ignores calls with different entities while a fetch is running. With the entity key fix from the comment above (entity key comparison + generation bump), this guard would only fire when the *same* entity set is already loading, which is the correct dedup behavior.

If you implement the entity key fix, this line becomes safe as-is — it would only skip when a duplicate fetch for the same entities is in progress.

ganeshrn (commented):
Review: clean extraction and caching pattern, one critical issue with cache identity.

abhikdps:
@coderabbitai review

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: fixed. The cache now tracks entity identity via `stringifyEntityRef`, so switching between a single-repo detail view and the global CI page invalidates and re-fetches instead of serving stale data. Also added a generation guard to discard in-flight writes after invalidation. ref commit: [9ea1596](https://github.com/ansible/ansible-backstage-plugins/pull/359/commits/9ea1596ec5b108eb62af69e5a3c5d902fa78ecf8)

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: same fix as the above. The cache now tracks entity identity via stringifyEntityRef, so switching between a single-repo detail view and the global CI page invalidates and re-fetches instead of serving stale data. Also added a generation guard to discard in-flight writes after invalidation. ref commit: [9ea1596](https://github.com/ansible/ansible-backstage-plugins/pull/359/commits/9ea1596ec5b108eb62af69e5a3c5d902fa78ecf8)

NilashishC:
## Code Review: feat: cache CI activity data across tab switches (#359)

**Verdict: NEEDS_CHANGES** | **Overall: 9.0/10**

| Lens | Score | Findings |
|---|---|---|
| Functionality | 8/10 | 2 Major |
| Security | 10/10 | 2 Nit |
| Quality | 9/10 | 2 Minor, 1 Nit |

### Major Findings

**1. [Functionality] `ciActivityCache.ts:93-100`** — Initial state skips full-page loading indicator

`startLoading` immediately sets `loading: false, fetchingMore: true` and notifies. The component derives `loading = !cacheState`, which becomes `false` after one render cycle. Users see a brief `<Progress />` flash then an empty table with "fetching more" indicator, instead of the original full-page loading that persisted until data started streaming.

<details>
<summary>Suggested fix</summary>

Set `loading: true` in the initial state, then switch to `loading: false, fetchingMore: true` when the first chunk request is sent. Update the component to use `cacheState?.loading ?? true` instead of `!cacheState`.

```diff
 // ciActivityCache.ts — initial state before fetch
   state = {
     rows: [],
-    loading: false,
-    fetchingMore: true,
+    loading: true,
+    fetchingMore: false,
     error: null,
     timestamp: Date.now(),
   };

 // RepositoriesCIActivityTab.tsx
-  const loading = !cacheState;
+  const loading = cacheState?.loading ?? true;
```
</details>

---

**2. [Functionality] `ciActivityCache.ts:35-40, 209-214`** — `invalidate()` nulls state without notifying subscribers

`notify()` has a guard `if (state)` that skips notification when state is null. After `invalidate()`, subscribed components retain stale data in local React state and never learn the cache was cleared.

<details>
<summary>Suggested fix</summary>

Update `Listener` type to accept `null`, remove the `if (state)` guard from `notify()`, and call `notify()` from `invalidate()`:

```diff
-type Listener = (state: CIActivityCacheState) => void;
+type Listener = (state: CIActivityCacheState | null) => void;

 function notify(): void {
-  if (state) {
-    for (const listener of listeners) {
-      listener(state);
-    }
+  for (const listener of listeners) {
+    listener(state);
   }
 }

 function invalidate(): void {
   generation += 1;
   state = null;
   entityKey = null;
   fetchPromise = null;
+  notify();
 }
```
</details>

---

### Minor Findings

**3. [Quality] `ciActivityCache.ts:216-221`** — `clear()` silently detaches all subscribers via `listeners.clear()`, inconsistent with `invalidate()`. Consider marking as test-only (`/** @internal */`) or notifying before clearing.

**4. [Quality] `ciActivityCache.ts` type** — `loading` field in `CIActivityCacheState` is dead code — always `false` in every state assignment, never read by consumers. Either remove it or use it meaningfully (see fix for #1).

### Nit

**5.** `normalizeGitHubStatus` (`ciActivityUtils.ts:55`) recreates the `valid` array on every call — hoist to module level.
**6.** `credentials: 'include'` (`ciActivityCache.ts:134`) is redundant with Backstage `FetchApi` defaults.

### Observations

- ✅ **CodeRabbit #2 (generation guard) — FIXED**: Generation counter correctly prevents stale writes.
- ✅ **CodeRabbit #3 (empty cachedEntities) — FIXED**: `cachedEntities.length > 0` check handles this.
- ⚠️ **CodeRabbit #1 (startLoading resolves immediately) — STILL PRESENT**: No runtime bug since callers use subscribe, but signature is misleading.
- Clean extraction of parsing logic with solid edge-case test coverage.
- The `cancelled` flag correctly prevents stale `getEntities()` completions.

### Path to 10/10

1. Use the `loading` field meaningfully — set `true` initially, switch to `false` + `fetchingMore: true` when fetching begins (+1 pt)
2. Notify subscribers when `invalidate()` clears state (+1 pt)
3. Mark `clear()` as test-only or notify before clearing listeners (+0.5 pts)
4. Remove or properly use the dead `loading` field (+0.5 pts)

---

*Assisted-by: Claude Code / Opus 4.6 (1M context) (Anthropic)*

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  state = {
+    rows: [],
+    loading: false,
+    fetchingMore: true,
+    error: null,
+    timestamp: Date.now(),
+  };
+  notify();
+
```
NilashishC, line 98: **[Major — Functionality]** Initial state sets `loading: false` immediately, so the component derives `loading = !cacheState = false` after one render cycle. Users see a brief `<Progress />` flash then an empty table with the "fetching more" indicator, instead of the original full-page loading that persisted until data started streaming.

```suggestion
  state = {
    rows: [],
    loading: true,
    fetchingMore: false,
    error: null,
    timestamp: Date.now(),
  };
```

Also update the component to derive loading from the cache state:
```typescript
const loading = cacheState?.loading ?? true;
```

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,224 @@
...
+function notify(): void {
+  if (state) {
+    for (const listener of listeners) {
+      listener(state);
+    }
+  }
+}
+
+function getState(): CIActivityCacheState | null {
```
NilashishC, line 38: **[Major — Functionality]** `notify()` guards on `if (state)`, so when `invalidate()` sets `state = null`, subscribers are never informed. Components retain stale data in local React state.

Update the `Listener` type to accept `null` and remove the guard:
```suggestion
function notify(): void {
  for (const listener of listeners) {
    listener(state);
  }
}
```
Also call `notify()` at the end of `invalidate()`.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,224 @@
...
+  entityKey = null;
+  fetchPromise = null;
+  listeners.clear();
+}
+
+export const ciActivityCache = {
+  getState,
+  subscribe,
+  startLoading,
+  setError,
```
NilashishC, line 221: **[Minor — Quality]** `clear()` calls `listeners.clear()` which silently detaches all subscribers — any mounted component stops receiving updates without knowing. `invalidate()` does not clear listeners, making the two methods inconsistent.

Consider either marking `clear()` as test-only (`/** @internal — test use only */`) or notifying listeners with `null` before clearing them.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
```diff
@@ -0,0 +1,224 @@
...
+import { CI_BATCH_CHUNK_SIZE, CI_PARALLEL_LIMIT } from './constants';
+
+const CI_ACTIVITY_CACHE_TTL_MS = 3 * 60 * 1000;
+
+export type CIActivityCacheState = {
+  rows: CIActivityRow[];
+  loading: boolean;
+  fetchingMore: boolean;
+  error: string | null;
+  timestamp: number;
```
NilashishC, line 14: **[Minor — Quality]** The `loading` field is dead code — it is set to `false` in every state assignment throughout this file and never read by any consumer. The component derives its loading state from `!cacheState` instead.

Either remove this field from the type, or use it meaningfully (set `true` initially, `false` after first chunk — see the suggestion on the initial state assignment).

NilashishC (dismissed):
9.0/10 — 2 Major findings on loading state and invalidation notification. See inline comments for details and suggested fixes. Full review in the summary comment above.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: Fixed. Initial state now sets loading: true, fetchingMore: false. Component derives `loading = cacheState?.loading ?? !cacheState`. Users see `<Progress />` until the first data chunk arrives.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: Fixed. Removed the if (state) guard, updated Listener type to accept null, and added notify() at the end of invalidate(). Subscribers are now informed when state is cleared.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: Marked `clear()` with `/** @internal - test use only */`. It's only called in test teardowns, not in production code - `invalidate()` is the production API and it now notifies listeners before clearing state.

plugins/self-service/src/components/GitRepositories/ciActivityCache.ts
abhikdps: Resolved by the fix for comment 1. The loading field is now used meaningfully: true initially, false after the first chunk arrives. The component reads it via `cacheState?.loading`.

PyContriver (approved):

PR #373, opened 2026-06-03
  [AAP-73727] fix: prevent task failures from OAuth token expiration during AAP job polling

PyContriver:
## Description

Fixes scaffolder task failures caused by OAuth token expiration during AAP job polling under concurrent load.

_The Problem:_
When multiple jobs queue in AAP, the scaffolder action's polling loop outlives the user's OAuth token, causing tasks to be marked "failed" even though the AAP jobs complete successfully.

_The Solution:_
 Remove polling from the scaffolder action and fetch job status on-demand from the frontend using a service account token. This makes AAP the single source of truth for job status, consistent with our existing catalog sync pattern.

_Key Changes:_
  - Scaffolder action launches jobs without blocking (returns immediately with job ID)
  - New backend API routes (GET /ansible/jobs/:jobId, POST /ansible/jobs/batch) expose job status using service account token
  - Frontend (RunTask, TaskList) fetches real-time AAP status and displays it alongside/instead of scaffolder status
  - Fully backwards compatible with existing tasks

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

## Checklist

- [x] Tests pass locally
- [x] Documentation updated

## Summary by CodeRabbit

* **New Features**
  * New job-status API endpoints: single-job GET and batch POST for AAP job status.
  * Optional non-blocking job-template launch (returns immediate pending job info).
  * UI: Task List and Run Task show AAP job status with live polling, links, and error fallback.

* **Tests**
  * Expanded coverage for launch flows, non-blocking behavior, single/batch status retrieval, batching, polling, and error cases.


PyContriver:
@coderabbitai review

api/openapi.yaml
```diff
@@ -352,6 +390,138 @@ paths:
...
+          description: Failed to fetch job status
+          content:
+            application/json:
+              schema:
+                allOf:
+                  - $ref: '#/components/schemas/ErrorResponse'
+                  - type: object
+                    properties:
+                      details:
+                        type: string
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

api/openapi.yaml
```diff
@@ -352,6 +390,138 @@ paths:
...
+            schema:
+              type: object
+              required:
+                - jobIds
+              properties:
+                jobIds:
+                  type: array
+                  items:
+                    type: integer
+                  description: Array of AAP job IDs to fetch status for
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.ts
```diff
@@ -780,6 +783,225 @@ export class AAPClient implements IAAPService {
...
+            url: `${this.getBaseUrl()}/execution/jobs/playbook/${jobID}/output`,
+          };
+        }
+      });
+
+      const batchResults = await Promise.all(promises);
+      batchResults.forEach(r => results.set(r.id, r));
+    }
+
+    return results;
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -954,5 +957,110 @@ export async function createRouter(options: {
...
+          .status(403)
+          .json({ error: 'Forbidden: insufficient permissions' });
+        return;
+      }
+
+      const jobId = parseInt(request.params.jobId, 10);
+      if (isNaN(jobId)) {
+        response.status(400).json({ error: 'Invalid job ID' });
+        return;
+      }
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -954,5 +957,110 @@ export async function createRouter(options: {
...
+          });
+        } else if (error.message?.includes('not configured')) {
+          response.status(503).json({
+            error: 'AAP service account not configured',
+          });
+        } else {
+          response.status(500).json({
+            error: 'Failed to fetch job status',
+            details: error.message,
+          });
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -954,5 +957,110 @@ export async function createRouter(options: {
...
+        return;
+      }
+
+      const validJobIds = jobIds.filter(
+        id => typeof id === 'number' && !isNaN(id),
+      );
+
+      if (validJobIds.length === 0) {
+        response.status(200).json({ jobs: {} });
+        return;
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/scaffolder-backend-module-backstage-rhaap/src/actions/aapLaunchJobTemplate.ts
```diff
@@ -45,7 +45,7 @@ export const launchJobTemplate = (ansibleServiceRef: IAAPService) => {
           normalized,
           'rhaap:launch-job-template',
         ) as LaunchJobTemplate;
-        jobResult = await ansibleServiceRef.launchJobTemplate(
+        jobResult = await ansibleServiceRef.launchJobTemplateNoWait(
           launchPayload,
           token,
         );
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/self-service/src/components/TaskList/TaskList.tsx
```diff
@@ -201,6 +212,51 @@ export const TaskList = () => {
...
+          ]),
+        );
+        setAapJobStatuses(statusMap);
+      } catch (err) {
+        // Silently fail - AAP status is optional enhancement
+      }
+    };
+
+    fetchAapStatuses();
+  }, [tasks, loading, discoveryApi]);
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/self-service/src/components/TaskList/TaskList.tsx
```diff
@@ -339,16 +395,43 @@ export const TaskList = () => {
...
+                            {(() => {
+                              const aapJobId = (task as any).output?.data?.id;
+                              const aapStatus = aapJobId
+                                ? aapJobStatuses.get(aapJobId)
+                                : null;
+
+                              const displayStatus =
+                                aapStatus && task.status === 'failed'
+                                  ? aapStatus.status
+                                  : task.status;
```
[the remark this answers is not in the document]
PyContriver: resolved in [41aece4](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/41aece4701360fb38c072a031ab8bd22c955ae64)

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -114,6 +117,28 @@ export async function createRouter(options: {
...
+   * Token should be passed from frontend in Authorization header or request body.
+   * Respects AAP RBAC by using user's token instead of service account.
+   */
+  async function getUserAAPToken(request: express.Request): Promise<string> {
+    // First check Authorization header
+    const authHeader = request.headers.authorization;
+    if (authHeader && authHeader.startsWith('Bearer ')) {
+      const token = authHeader.substring(7);
+      if (token) return token;
+    }
```
abhikdps: this would extract the backstage JWT and not the AAP token, right? is this intentional? I think we need the AAP token and not the backstage one.

PyContriver: @abhikdps Your analysis is correct. To address this, we're now passing the AAP OAuth token via a custom AAP-Token header from the frontend (instead of using the Authorization header which contains the Backstage JWT). This follows the same pattern as the existing X-Github-Token header that we use for github. The change is in this commit: [9134320](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/9134320ee6933212c28a921d03e8556fa868b732)

plugins/backstage-rhaap-common/src/AAPClient/utils/jobTemplateHelpers.ts
```diff
@@ -0,0 +1,111 @@
...
+  limit?: string;
+  verbosity?: number;
+  job_slice_count?: number;
+  timeout?: number;
+  diff_mode?: boolean;
+  job_tags?: string;
+  skip_tags?: string;
+  extra_vars?: object | string;
+  credentials?: number[];
+} {
```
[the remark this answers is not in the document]
PyContriver: addressed in [436b2a1](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/436b2a19177ca34e43f41f0b9f71823cbfc649ab)

plugins/self-service/src/components/RunTask/RunTask.tsx
```diff
@@ -140,6 +155,70 @@ export const RunTask = () => {
...
+      cancelled = true;
+      if (intervalId) {
+        clearInterval(intervalId);
+      }
+    };
+  }, [task, completed, aapJobStatus?.status, discoveryApi, taskId, aapAuth]);
+
   const canStartOver = canRead && canCreateTask;
   const showStartOver = canStartOver;
   const isStartOverDisabled =
```
abhikdps, line 145: **P0 — Polling stale closure / unnecessary teardown-rebuild cycles**

`aapJobStatus?.status` is in the dependency array, so every status transition (`pending` → `running` → `successful`) tears down this entire effect and re-creates it — clearing the interval, calling `fetchAapStatus()` eagerly, then conditionally restarting the interval. This works by accident (cleanup clears the old interval before the new one starts), but:

1. Each re-entry fires an **extra immediate fetch** before re-evaluating `isPending` — so every status change triggers a redundant request.
2. `isPending` is computed from render-time state, not the latest async result — there's a window where the fetch resolves to a terminal status, state updates, the effect re-runs, but `isPending` on the *new* render correctly sees the terminal status and doesn't set an interval. The race happens if two fetches are in flight during the teardown/rebuild.
3. This pattern is fragile for anyone modifying it later — adding another dep to the array or changing the timing could create leaked intervals.

**Suggested fix:** Remove `aapJobStatus?.status` from the dependency array. Use a ref (`useRef`) to track the latest status, and check for terminal status *inside* the interval callback. Clear the interval from within when a terminal status is reached:

```tsx
const statusRef = useRef<string | null>(null);

useEffect(() => {
  // ... guard checks ...
  const fetchAapStatus = async () => {
    // ... fetch logic ...
    statusRef.current = jobStatus.status;
    if (['successful','failed','error','canceled'].includes(jobStatus.status)) {
      clearInterval(intervalId);
    }
  };
  fetchAapStatus();
  const intervalId = setInterval(fetchAapStatus, 5000);
  return () => clearInterval(intervalId);
}, [task, completed, discoveryApi, taskId, aapAuth]);
```

This way the effect runs once, the interval self-terminates on completion, and there are no teardown/rebuild cycles on every status change.

plugins/self-service/src/components/TaskList/TaskList.tsx
```diff
@@ -201,6 +212,70 @@ export const TaskList = () => {
...
+    const abortController = new AbortController();
+
+    const fetchAapStatuses = async () => {
+      try {
+        const baseUrl = await discoveryApi.getBaseUrl('catalog');
+        const response = await fetch(`${baseUrl}/ansible/jobs/batch`, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
```
abhikdps: **P1 — Missing `AAP-Token` header — this endpoint will silently fail every time**

The batch request doesn't send the `AAP-Token` header and doesn't call `aapAuth.getAccessToken()`. The router's `getUserAAPToken()` helper will throw `"AAP OAuth token required"`, the response will be non-200, and the `if (!response.ok) return` below silently swallows the error.

This means the entire batch AAP status feature on the task list page is broken — it will never successfully fetch statuses.

**Fix:** Import `rhAapAuthApiRef`, call `getAccessToken()`, and set the `AAP-Token` header — same pattern as `RunTask.tsx`:

```tsx
const aapAuth = useApi(rhAapAuthApiRef);
// ...
const fetchAapStatuses = async () => {
  const aapToken = await aapAuth.getAccessToken();
  const response = await fetch(`${baseUrl}/ansible/jobs/batch`, {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
      'AAP-Token': aapToken,
    },
    // ...
  });
};
```

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -993,5 +1015,185 @@ export async function createRouter(options: {
     },
   );
 
+  /**
```
abhikdps: **P1 — Stale comment: says "service account token" but code uses user's OAuth token**

This comment says `"Uses service account token from config"` but the implementation below calls `getUserAAPToken(request)` which extracts the user's AAP OAuth token from the `AAP-Token` header. Same issue exists on the batch endpoint comment further down.

Please update both handler comments to reflect the actual implementation.

abhikdps (commented):
The architecture is sound — fire-and-forget scaffolder with live AAP status via frontend is the right pattern. The `AAP-Token` header separation, `buildLaunchPayload` extraction, and helper tests are all good fixes since last round.

**Two blockers remain (P0 + P1), plus stale documentation across the PR.**

See inline comments for details.

api/openapi.yaml
```diff
@@ -425,6 +463,138 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
 
+  /ansible/jobs/{jobId}:
+    get:
+      operationId: getJobStatus
+      summary: Get AAP job status by job ID
+      description: |
```
abhikdps: **P1 — OpenAPI description still references service account token**

This description says:
> Uses the service account token from configuration, not the user's OAuth token.

But the code now uses the user's OAuth token via the `AAP-Token` header. The batch endpoint description (line ~590) has the same issue. Please update both to match the current implementation — consumers reading the API spec will be misled about the auth model.

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -993,5 +1015,185 @@ export async function createRouter(options: {
...
+        );
+        // Check for 404 or 403 (not found / no access)
+        if (
+          error.message?.includes('404') ||
+          error.message?.includes('403') ||
+          error.message?.toLowerCase().includes('not found')
+        ) {
+          response.status(404).json({
+            error: 'Job not found or insufficient permissions',
+            id: jobId,
```
abhikdps: **P2 — Error detection via string matching is brittle**

```ts
error.message?.includes('404') ||
error.message?.includes('403') ||
error.message?.toLowerCase().includes('not found')
```

This will false-positive on any error message that happens to contain "403" or "404" as a substring (e.g., `"Processed 404 records"`, `"Port 4043 connection refused"`). 

Consider using typed errors (e.g., a custom `AAPError` class with a `statusCode` property) or checking the HTTP status code from the AAP response directly, rather than parsing error message strings.

plugins/catalog-backend-module-rhaap/src/router.ts
```diff
@@ -993,5 +1015,185 @@ export async function createRouter(options: {
...
+
+      // Validate all entries have taskId and jobId
+      const invalidEntries = jobRequests.filter(
+        req =>
+          !req.taskId ||
+          typeof req.taskId !== 'string' ||
+          typeof req.jobId !== 'number' ||
+          !Number.isInteger(req.jobId) ||
+          req.jobId <= 0,
+      );
```
abhikdps: **P2 — No upper bound on batch size**

The batch endpoint accepts any number of job IDs. While `getJobStatusBatch` processes them in batches of 10, a malicious or buggy client could send thousands of IDs, causing the backend to make thousands of sequential requests to AAP.

Consider adding a max (e.g., 50 or 100) and returning 400 if exceeded:
```ts
if (jobRequests.length > 100) {
  response.status(400).json({ error: 'Maximum 100 jobs per batch request' });
  return;
}
```

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.test.ts
```diff
@@ -4269,4 +4269,469 @@ describe('AAPClient', () => {
...
+          results: [{ id: 123, name: 'Test Template' }],
+        }),
+      };
+
+      const launchResponse = {
+        ok: true,
+        status: 201,
+        json: jest.fn().mockResolvedValue({
+          job: 456,
+        }),
```
abhikdps, line 4292: **Nit — Test name still says "service account token"**

This test name (`'should fetch job status using service account token'`) and a few others below (`'should throw error when service account token not configured'` at ~line 4359 and ~4524) still reference the old service-account approach. The code now uses the user's OAuth token. Please update the test descriptions to match.

audgirka (commented):
The batch endpoint's OpenAPI spec defines `jobIds: integer[]` but the router expects `jobs: Array<{ taskId, jobId }>`. API clients generated from the spec would send the wrong payload. Also the single job endpoint uses `parseInt` which accepts values like `123abc` — consider using `Number()` instead for stricter validation. Otherwise the approach looks solid, the launch/poll split is the right fix for the token expiry issue.

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.test.ts
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

plugins/catalog-backend-module-rhaap/src/router.ts
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

api/openapi.yaml
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

plugins/catalog-backend-module-rhaap/src/router.ts
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

plugins/self-service/src/components/TaskList/TaskList.tsx
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

plugins/self-service/src/components/RunTask/RunTask.tsx
PyContriver: Addressed this as a part of [ad456f2](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/ad456f20d2f7e946886c2662f852104f087cce03) and [a8caa6d](https://github.com/ansible/ansible-backstage-plugins/pull/373/commits/a8caa6d333f838c4b85ecb09e00dca2625c4e268)

ganeshrn (commented):
## Architecture concern: user token polling doesn't solve the root problem

The PR description states the goal: *"tasks marked 'failed' even though the AAP jobs complete successfully"* when OAuth tokens expire during long polls.

The current implementation moves polling from the backend scaffolder action to the frontend, passing the user's AAP OAuth token via `AAP-Token` header on every request. This solves access token expiration (frontend calls `getAccessToken()` which auto-refreshes), but **doesn't solve the root scenario**:

### What still breaks

1. **User closes the tab / laptop sleeps.** Polling stops. When they return, they see stale scaffolder "failed" status until the page reloads. For long-running AAP jobs (the exact use case), users won't sit and watch.

2. **Refresh token expiration.** AAP refresh tokens have their own TTL (configurable by AAP admin, typically hours). A job that outlasts the refresh token leaves the user with no recovery path except re-login.

3. **AAP token transit exposure.** The user's AAP OAuth token travels from frontend → Backstage backend on every poll via a custom header. Any request-logging middleware or proxy between them captures a live credential for an external system. The `getUserAAPToken()` body-fallback path compounds this.

### Proposed approach: service account token + task ownership validation

Use the service account token already configured at `ansible.rhaap.token` for job status lookups, and enforce authorization at the Backstage layer instead of delegating to AAP RBAC:

```
Frontend polls backend → Backend uses service account token to query AAP
                        → Backend queries scaffolder task store to verify
                          requesting user owns the task that launched the job
                        → Returns status only if ownership check passes
```

**What this fixes:**
- No user token needed for job status → no expiration problem at all
- Backend can check job status even when user is offline (enables future webhook/SSE push)
- No AAP credentials transit through frontend requests
- `taskId` parameter becomes a real authorization check, not just a log annotation

**What's needed:**
1. `getJobStatus` / `getJobStatusBatch` use `this.token` (service account) instead of requiring a user token parameter
2. Router validates task ownership: query scaffolder task store, verify `taskId` exists and was created by the requesting user, verify the job ID matches the task's output
3. Remove `AAP-Token` header requirement from both endpoints
4. Remove `getUserAAPToken()` helper and body-token fallback

**Trade-off:** The service account may see jobs the user couldn't see directly in AAP. The task ownership check compensates: users can only query jobs *they launched through the Portal*, which is a stricter scope than AAP RBAC in practice (user can't query arbitrary AAP job IDs even if AAP would allow it).

This is a required change in my view — the current approach doesn't reliably fix the stated problem for the primary use case (long-running jobs where users navigate away).

ganeshrn (commented):
## Review: Architecture concern + code issues

### Architecture: user token polling doesn't solve the root problem

The PR goal: *"tasks marked 'failed' even though the AAP jobs complete successfully"* when OAuth tokens expire during long polls.

The current implementation moves polling from the backend scaffolder action to the frontend, passing the user's AAP OAuth token via `AAP-Token` header on every request. This solves access token expiration (frontend calls `getAccessToken()` which auto-refreshes), but doesn't solve the root scenario:

**What still breaks:**

1. **User closes the tab / laptop sleeps.** Polling stops. When they return, they see stale scaffolder "failed" status until the page reloads. For long-running AAP jobs (the exact use case), users won't sit and watch.

2. **Refresh token expiration.** AAP refresh tokens have their own TTL. A job that outlasts the refresh token leaves the user with no recovery path except re-login.

3. **AAP token transit exposure.** The user's AAP OAuth token travels from frontend to Backstage backend on every poll via a custom header. Any request-logging middleware or proxy captures a live credential for an external system.

**Proposed approach: service account token + task ownership validation**

Use the service account token already configured at `ansible.rhaap.token` for job status lookups, and enforce authorization at the Backstage layer:

- `getJobStatus` / `getJobStatusBatch` use the service account token instead of requiring a user token parameter
- Router validates task ownership: query scaffolder task store, verify `taskId` exists and was created by the requesting user, verify the job ID matches the task's output
- Remove `AAP-Token` header requirement and `getUserAAPToken()` helper

Frontend polls backend → Backend uses service account token to query AAP
                        → Backend queries scaffolder task store to verify
                          requesting user owns the task that launched the job
                        → Returns status only if ownership check passes
What this fixes:
```
No user token needed for job status → no expiration problem at all
Backend can check job status even when user is offline (enables future webhook/SSE push)
No AAP credentials transit through frontend requests
taskId parameter becomes a real authorization check, not just a log annotation
```
**Trade-off:** The service account may see jobs the user couldn't see directly in AAP. The task ownership check compensates -- users can only query jobs they launched through the Portal, which is stricter than AAP RBAC in practice.

---

### Critical issues

**1. `router.ts` batch endpoint -- no upper bound on `jobs` array size**

No max length check on the batch array. A client could send thousands of job IDs, causing the backend to fire thousands of API calls to AAP (batched 10 at a time via `BATCH_SIZE` in `getJobStatusBatch`, but total volume is unbounded).

Fix: Add a cap (e.g., 50 or 100) and return 400 if exceeded.

**2. `router.ts` -- `taskId` is required but never validated for ownership**

The OpenAPI spec describes `taskId` as "for ownership validation." The code requires it but only uses it in a log message. No actual check verifies the requesting user owns the scaffolder task. This becomes the core of the proposed architecture fix above.

**3. `router.ts:130-134` -- `getUserAAPToken()` body-token fallback is undocumented**

The OpenAPI spec documents only the `AAP-Token` header, but the code also accepts a token from `request.body.token`. This creates an undocumented auth path where tokens could be captured by request-body logging middleware. This goes away entirely with the proposed service account approach.

**4. `router.ts:1065-1090` -- error routing via exact string matching**

Error-to-HTTP-status mapping relies on matching exact error message strings from AAPClient (`'Insufficient privileges. Please contact your administrator.'`, `'Failed to fetch data.'`). If AAPClient changes any message text, the mapping silently breaks and falls through to 500. The code acknowledges this limitation in a comment.

Fix: Throw typed errors from AAPClient (e.g., `class AAPNotFoundError`) with a `statusCode` property, or use error codes instead of message matching.

---

### Informational issues

**5. `RunTask.tsx:158` -- `statusRef` is set but never read**

`const statusRef = useRef<string | null>(null)` is assigned on each poll but never consumed elsewhere. Dead code.

**6. `RunTask.tsx` / `TaskList.tsx` -- `(task as any).output?.data?.id` type assertions**

Used in 4 places across both components. Bypasses type checking entirely. If `ScaffolderTask` doesn't have `output` in its type, extend the interface rather than casting to `any`.

**7. `RunTask.tsx` -- no polling backoff**

Polling interval is hardcoded at 5000ms with no backoff. For jobs running hours, that's 720 calls/hour indefinitely. Consider starting at 5s and doubling up to 30s max. (Terminal status stop is already implemented correctly.)

**8. `TaskList.tsx` -- `getStatusIcon` may not handle AAP status strings**

`displayStatus` can be an AAP status (`pending`, `running`, `successful`, `canceled`) passed to `getStatusIcon()`. If that function only handles Backstage scaffolder statuses (`completed`, `processing`, `open`, `failed`), unrecognized AAP statuses render without an icon or hit an unexpected default.

ganeshrn (commented):
test

ganeshrn (changes requested):
Requesting changes based on the architecture concern and critical issues outlined in the review comment: https://github.com/ansible/ansible-backstage-plugins/pull/373#issuecomment-4665926095

**Required changes:**

1. **Use service account token instead of user's AAP OAuth token for job status lookups.** The current approach doesn't solve the root problem (user closes tab / refresh token expires). Use `ansible.rhaap.token` and enforce authorization via task ownership validation at the Backstage layer.

2. **Validate task ownership.** Query scaffolder task store to verify the requesting user owns the task that launched the job. `taskId` is currently logged but never validated.

3. **Add batch size limit.** The `/ansible/jobs/batch` endpoint accepts unbounded arrays. Cap at 50-100 and return 400 if exceeded.

4. **Remove `getUserAAPToken()` body-token fallback.** Undocumented auth path that contradicts the OpenAPI spec and exposes tokens in request body logs.

5. **Replace error message string matching with typed errors.** Error-to-HTTP-status mapping in the job status endpoint relies on exact string matches from AAPClient. Use typed error classes with status codes instead.

audgirka (approved):
Looks good. The launch/poll split is the right fix for the token expiry issue. Service account token for status polling makes sense since user tokens can expire between sessions.

UI already filters tasks by ownership (non-admins only see their own tasks), so the TODO for backend task ownership validation is non-blocking — defense-in-depth for later. The admin token is only used for read-only status checks, not launching.

One thing to track as a follow-up: implement the validateTaskOwnership TODO so the API endpoint doesn't rely solely on the UI filter for access control.

ganeshrn:
## Re-Review: PR #373 v2 (post architecture rework)

Significant improvements since v1. The core fix moved from frontend-token-polling to service-account-token polling in the scaffolder action, which solves the root problem. Several issues from the previous review are resolved. New issues surfaced from the rework.

### Previous findings status

| # | v1 Finding | v2 Status |
|---|---|---|
| 1 | Batch endpoint no size limit | **Resolved** - router endpoints removed |
| 2 | taskId not validated for ownership | **Resolved** - router endpoints removed |
| 3 | getUserAAPToken body-token fallback | **Resolved** - removed entirely |
| 4 | Error string matching in router | **Resolved** - router endpoints removed |
| 5 | statusRef dead code in RunTask | **Resolved** - RunTask.tsx removed from diff |
| 6 | (task as any) type assertions | **Resolved** - replaced with `TaskWithAAPJob` type |
| 7 | No polling backoff | **Changed** - polling now in backend scaffolder action |
| 8 | getStatusIcon AAP statuses | **Moot** - dead code, see finding #3 |

### Critical

**1. `aapLaunchJobTemplate.ts` -- polling loop has no timeout or max iterations**

```typescript
while (
  currentStatus &&
  !['successful', 'failed', 'error', 'canceled'].includes(currentStatus)
) {
  await new Promise(resolve => setTimeout(resolve, 5000));
  const statusUpdate = await ansibleService.getJobStatus(jobResult.id, pollingToken);
  currentStatus = statusUpdate.status?.toLowerCase();
}
```

If AAP returns an unexpected status (e.g., `"waiting"`, `"new"`) or the job gets stuck, this loops forever. The scaffolder task never completes, consuming a backend worker indefinitely.

Fix: Add max poll count or timeout. E.g., `const MAX_POLLS = 720;` and break with error after limit (720 * 5s = 1 hour).

**2. `aapLaunchJobTemplate.ts` -- silent fallback to user token defeats the fix**

```typescript
const pollingToken = serviceToken || token;
```

If `ansible.rhaap.token` is not configured, this silently falls back to the user's token, which is exactly the token expiration problem this PR fixes. Operators won't know the fix isn't active.

Fix: Log a warning when falling back: `if (!serviceToken) { logger.warn('ansible.rhaap.token not configured - falling back to user token for polling. Long-running jobs may fail due to token expiry.'); }`

**3. `TaskList.tsx` -- remove dead AAP status code**

`TaskList.tsx` calls `POST /ansible/jobs/batch` but this endpoint does not exist in `router.ts` (verified on both `main` and `AAP-73727` branches). The call silently 404s, so the AAP status column never populates.

This code is also unnecessary. The scaffolder action now polls server-side and writes the final AAP job status into the task output. The TaskList already displays task status from the scaffolder API, which includes the correct AAP result once polling completes. There is no gap to fill.

Remove the dead code from this PR:
- `TaskList.tsx`: remove `aapJobStatuses` state, `fetchAapStatuses` effect, `TaskWithAAPJob` type, `discoveryApiRef` import, and AAP status display logic in the table render
- `TaskList.test.tsx`: remove `mockDiscoveryApi` and related test API provider entries
- `RunTask.test.tsx`: remove `mockDiscoveryApi` and related test API provider entries

### Informational

**4. `AAPClient.ts` `launchJobTemplateNoWait` -- misleading log messages**

```typescript
this.logger.info(`Start executing job template.`);
// ...
this.logger.info(`Waiting for result of the executed job template.`);
```

The `NoWait` variant logs "Waiting for result" but doesn't wait. These were copied from the blocking `launchJobTemplate` method.

---

```
Re-Review Summary: 4 issues (3 critical, 1 informational)

CRITICAL:
  [1] Unbounded polling loop - no timeout, can hang forever
  [2] Silent fallback to user token defeats the fix when service token missing
  [3] Remove dead TaskList AAP status code - endpoint doesn't exist, code unnecessary

INFORMATIONAL:
  [4] Misleading "Waiting for result" log in NoWait method

Architecture: Service-account-token polling in the scaffolder action is the right
approach and solves the stated problem.
```

ganeshrn:
## Review v3: All previous findings resolved

Latest commit `fe695e0` (fix: review comments) addresses all four findings from the previous review:

| # | Finding | Resolution |
|---|---|---|
| 1 | Unbounded polling loop | **Fixed** -- `MAX_POLLS = 720` (1 hour timeout), throws error with last status on expiry |
| 2 | Silent fallback to user token | **Fixed** -- `logger.warn` when `ansible.rhaap.token` not configured |
| 3 | Dead TaskList AAP status code | **Fixed** -- all AAP status code removed from TaskList (batch endpoint calls, `aapJobStatuses` state, `fetchAapStatuses` effect, `TaskWithAAPJob` type) |
| 4 | Misleading "Waiting for result" log in NoWait | **Fixed** -- moved to scaffolder action's blocking path where it actually waits |

```
Pre-Landing Review v3: No issues found.
```

Architecture is sound: scaffolder action launches with user token (AAP RBAC enforced), polls with service account token (no expiration), 1-hour timeout guard. `buildLaunchPayload` extraction is a clean refactor with thorough test coverage.

ganeshrn (approved):

audgirka (approved):

PR #388, opened 2026-06-09
fix: EE catalog starred filter not updating table on unstar

abhikdps:
## Summary

Fixes the starred filter behavior and pagination issues on the Execution Environments catalog page.

### Starred filter fixes
- **Unstarring entities while in "starred" filter mode** now correctly updates the table by re-issuing `EntityUserFilter.starred()` with the updated entity refs
- **Unstarring the last starred entity** now switches back to showing all entities instead of showing the empty "No Execution Environment definition files, yet..." page
- **Prevents unmount race condition** where the `UserListPicker` component would unmount (due to `hasEntities` becoming false) before its internal auto-switch from "starred" to "all" could complete
- Added defensive default (`= new Set<string>()`) for `starredEntities` destructuring to handle cases where the hook returns `undefined`

### Pagination fixes
- Derive `page` directly from `offset/limit` instead of local `useState` — keeps the page indicator in sync with server-side pagination state
- Removed the `useEffect` that was clamping offset, which caused unnecessary re-renders
- Fixed `onPageChange` to update offset via `setOffset(p * limit)` instead of setting local state
- Fixed `onRowsPerPageChange` to reset offset to 0 when page size changes
- Removed `setPage(0)` calls from filter change handlers (no longer needed)
- Added type annotation for `getEntityFacets` response

### Test updates
- Enhanced `Table` stub mock with `page`, `totalCount`, `onPageChange`, `onRowsPerPageChange`, and `options` props
- Updated pagination test to verify page is derived from offset/limit without calling `setOffset`
- Added test for rows-per-page change resetting offset to 0

## Test plan

- [x] Navigate to EE catalog page
- [x] Star several entities, switch to "Starred" filter — only starred entities shown
- [x] Unstar an entity — table updates immediately showing remaining starred entities
- [x] Unstar the last entity — table switches to showing all entities (not the empty state)
- [x] Verify URL updates from `filters[user]=starred` to no user filter
- [x] Pagination: navigate to page 2, verify correct entities shown
- [x] Change rows per page — verify page resets to first page
- [x] Apply owner/tag filters — verify pagination resets correctly
- [x] Navigate away and back — page loads correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pagination handling and offset management for filtered catalog lists.
  * Enhanced starred-filter synchronization when switching between views.
  * Better state persistence when modifying filters and page size.

* **Tests**
  * Expanded test coverage for pagination interactions and starred-filter behavior.

### DIFF
--- a/plugins/self-service/src/components/ExecutionEnvironments/catalog/CatalogContent.test.tsx
+++ b/plugins/self-service/src/components/ExecutionEnvironments/catalog/CatalogContent.test.tsx
@@ -22,9 +22,35 @@ jest.mock('@backstage/core-components', () => {
   const actual = jest.requireActual('@backstage/core-components');
   return {
     ...actual,
-    Table: ({ title, data = [], columns = [] }: any) => (
+    Table: ({
+      title,
+      data = [],
+      columns = [],
+      page,
+      totalCount,
+      onPageChange,
+      onRowsPerPageChange,
+      options,
+    }: any) => (
       <div data-testid="stubbed-table">
         <div data-testid="stubbed-table-title">{title}</div>
+        <div data-testid="stubbed-table-page">{page ?? ''}</div>
+        <div data-testid="stubbed-table-total">{totalCount ?? ''}</div>
+        <div data-testid="stubbed-table-pagesize">
+          {options?.pageSize ?? ''}
+        </div>
+        {onPageChange && (
+          <button
+            data-testid="stubbed-table-next-page"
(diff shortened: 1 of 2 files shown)


abhikdps:
@coderabbitai review

abhikdps:
@coderabbitai review

audgirka (approved):
LGTM. Clean fix for the starred filter re-render. Good use of prevStarredSizeRef for edge case when last star is removed.

PR #402, opened 2026-06-19
[AAP-79854] fix: resolve superuser permission denial from auth token race condition

abhikdps:
## Problem

AAP superusers who are members of `group:default/aap-admins` are denied all ansible permissions on deployed RHDH instances, even when the RBAC config lists the group in `superUsers`. The Self-Service sidebar items (EE Builder, Collections, Git Repositories, Templates, History) are hidden, and the `/rbac` admin page is inaccessible.

This happens because `signInWithCatalogUser` builds the token's ownership refs (`ent` claim) from `entity.relations`. For newly created users, the catalog hasn't stitched the `memberOf` relation to `group:default/aap-admins` yet — so the token is issued without it. The RBAC plugin then can't match the user to the `superUsers` list via group membership, and falls through to Casbin policy lookup where no ansible permission policies exist.

Only `user:default/admin` (matched by exact `userEntityRef`) gets the superUser bypass. All other superusers are denied.

## Description

- Fix race condition where `signInWithCatalogUser` issues tokens before catalog stitches `memberOf` relations, causing superusers to be denied all ansible permissions (Self-Service pages hidden)
- Replace with `findCatalogUser` + `issueTokenWithOwnership` that checks the `aap.platform/is_superuser` annotation directly (available immediately, no stitching dependency) to force `group:default/aap-admins` into the `ent` claim
- Works across both RHDH 1.9 and RHDH 1.10+ (RBAC v7.12.5)

## Related Issues

Related JIRA: [AAP-79854](https://redhat.atlassian.net/browse/AAP-79854)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Existing resolver tests updated to use `findCatalogUser` + `issueToken`
- [x] New test: superuser gets `group:default/aap-admins` in token via annotation
- [x] New test: superuser on first login (race condition scenario) gets correct token
- [x] New test: non-superuser with stitched group relations gets correct token
- [x] Verify on deployed RHDH instance that Self-Service sidebar items are visible

## Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **Improvements**
  * Updated catalog-based sign-in to issue tokens directly from catalog user entity data, including ownership entries derived from the user’s stitched group memberships.
  * Superuser logins now include the admin group entry in the token’s ent claims.
  * Consistent token-issuance behavior across username matching and new-user sign-in flows.

* **Tests**
  * Expanded resolver tests to validate issued token claims (including superuser/admin inclusion) and memberOf-derived ent entries.

### DIFF
--- a/plugins/auth-backend-module-rhaap-provider/src/resolvers.test.ts
+++ b/plugins/auth-backend-module-rhaap-provider/src/resolvers.test.ts
@@ -6,6 +6,30 @@ import {
 } from '@backstage/plugin-auth-node';
 import { AAPAuthSignInResolvers } from './resolvers';
 
+function mockUserEntity(
+  name: string,
+  opts?: {
+    isSuperuser?: boolean;
+    memberOfGroups?: string[];
+  },
+) {
+  const annotations: Record<string, string> = {};
+  if (opts?.isSuperuser !== undefined) {
+    annotations['aap.platform/is_superuser'] = String(opts.isSuperuser);
+  }
+  const relations = (opts?.memberOfGroups ?? []).map(g => ({
+    type: 'memberOf',
+    targetRef: `group:default/${g}`,
+  }));
+  return {
+    apiVersion: 'backstage.io/v1alpha1',
+    kind: 'User',
+    metadata: { name, namespace: 'default', annotations },
+    spec: {},
+    relations,
+  };
+}
+
 global.fetch = jest.fn();
 
 const mockDiscovery = {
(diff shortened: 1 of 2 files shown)


audgirka (approved):
Reviewed. The approach is sound — `issueTokenWithOwnership` bypasses the catalog stitching race by reading the `aap.platform/is_superuser` annotation directly and injecting `group:default/aap-admins` into the ownership refs.

Confirmed that `omitIdentityTokenOwnershipClaim: true` (app-config.yaml:83) does not affect this fix — ownership refs flow through `UserInfoService` regardless, and the RBAC plugin reads from there, not the JWT.

One note: `issueTokenWithOwnership` correctly returns the full result from `ctx.issueToken()` rather than destructuring `{ token }` — this is important because `UserInfoService` needs the complete result object to persist ownership refs.

abhikdps:
@coderabbitai review

PR #367, opened 2026-05-28
feat: multi-org support — namespace scoping, logout sync, entity providers

craig-br:
## Summary

Adds multi-org support for AAP organizations in the portal catalog, enabling deployments with 2+ AAP orgs to have proper namespace isolation, entity scoping, and logout synchronization.

Supersedes #328 (branch renamed from `test/multi-org` to `feat/multi-org-prototype`).

### Key design decisions
- **Single-org = zero disruption**: When only 1 org is configured, entities stay in `default` namespace — no behavior change on upgrade
- **Multi-org activates scoping**: When 2+ orgs configured, `getEffectiveNamespace()` maps org names to Backstage-safe namespaces (e.g., `Default` → `aap-default`)
- **`aap-admins` always in `default`**: Cross-org superuser group stays in the default namespace for consistent RBAC
- **Namespace collision detection**: Validates that distinct org names don't map to the same namespace
- **Job template titles**: No `[OrgName]` suffix — org attribution via `ansible.com/organization` annotation driving a UI chip

### Changes

| Area | Files | What |
|------|-------|------|
| Namespace helpers | `helpers.ts`, `helpers.test.ts` | `formatNameSpace()`, `getEffectiveNamespace()`, `validateNamespace()`, collision detection |
| Entity providers | `AAPEntityProvider.ts`, `AAPJobTemplateProvider.ts` + tests | Org namespace scoping, cross-namespace membership refs, org annotations, collision validation |
| Auth — FR-011 | `AAPClient.ts` + tests | `fetchProfile` distinguishes network errors from 401s — prevents transient failures from logging out users |
| Logout sync | `AAPLogoutHandler` + tests, `authenticator.ts` + tests | AAP session revocation on Backstage signout; refresh-failure detection |
| Config | `config.ts` | Org-empty guard — logs warning and syncs nothing when no orgs configured (SC-009) |
| Entity parsers | `entityParser.ts`, `dynamicJobTemplate.ts` | Conditional `ansible.com/organization` annotations, namespace passthrough |
| E2E tests | `multi-org-*.spec.ts` (3 files), `backstage-api.ts` | Template visibility, catalog API validation, UI entity page verification |

## Test plan
- [x] `yarn tsc` — no type errors
- [x] Unit tests — helpers, AAPClient, AAPLogoutHandler, AAPJobTemplateProvider, AAPEntityProvider, entityParser (all passing)
- [x] Job template title test fixed to match plan design (no org suffix)
- [ ] Review

[...]

PR #408, opened 2026-06-23
[ANSTRAT-912] feat(catalog): add multi-org namespace scoping, logout sync, and hardening

audgirka:
### Description

Adds multi-org support for AAP organizations in the portal catalog, enabling deployments with 2+ AAP orgs to have namespace isolation, entity scoping, and logout synchronization. Supersedes #367.

Key changes from prototype + hardening:
- `getEffectiveNamespace()` maps org names to Backstage-safe namespaces (single-org = zero disruption)
- `formatNameSpace()` hardened: strict `[a-z0-9-]` charset, underscore→hyphen, 63-char truncation
- Config-time namespace validation rejects invalid org names at startup
- Org name trim in URLSearchParams prevents `+prefix` bug in AAP queries
- AAPLogoutHandler: WeakSet guard prevents double-mount race condition
- CreateTask: passes full `templateRef` with namespace to `getTemplateParameterSchema`

### Related Issues

Supersedes #367
Related JIRA: [ANSTRAT-912](https://redhat.atlassian.net/browse/ANSTRAT-912)

### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### Testing

- Unit tests pass (805 catalog-backend, 10 logout handler)
- Multi-org tested locally with 3 AAP orgs (Default, Engineering, SecOps) — 29 templates synced with correct namespaces
- Single-org backward compatibility verified (entities stay in `default` namespace)
- E2E Playwright specs added for multi-org catalog API, template visibility, UI entity page
- Lint clean

### Checklist

- [x] Tests pass locally
- [ ] Documentation updated

## Summary by CodeRabbit

* **New Features**
  * Added multi-organization support across Catalog sync, job template sync/visibility, and UI, including org-scoped namespaces and organization-related annotations.
  * Implemented template redirects to self-service catalog paths and enhanced logout with a dedicated handler plus safer logout button flow.

* **Improvements**
  * Stricter namespace normalization/validation and consistent org-scoped group/user references.
  * Improved RH AAP authentication/session error handling and more reliable profile fetching.
  * Updated logout redirection behavior to handle trailing base URLs.

* **Configuration**
  * Switched dev SQLite to persistent file-backed storage and expanded ignored artifacts.
  * Allowed catalog `orgs` configuration to accept a single value or a list.

* **Tests**
  * Added/expanded Playwright multi-org e2e coverage and extended auth/logout unit/component tests.


deploy/index.json
abhikdps, line 1: this file can be removed

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.ts
```diff
@@ -1027,24 +1040,39 @@ export class AAPClient implements IAAPService {
...
-      const endPoint = 'api/gateway/v1/me/';
-      response = await this.executeGetRequest(endPoint, token);
+      response = await fetch(url, {
+        method: 'GET',
+        dispatcher: this.proxyAgent,
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bearer ${token}`,
+        },
+      });
```
abhikdps, line 1056: why change this to use fetch and not use executeGetRequest method?

plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.ts
```diff
@@ -127,6 +130,23 @@ export class AAPJobTemplateProvider implements EntityProvider {
...
       throw new NotFoundError('Not initialized');
     }
+
+    if (!this.orgs || this.orgs.length === 0) {
+      this.logger.warn(
+        `[${AAPJobTemplateProvider.pluginLogName}]: No orgs configured in catalog.providers.rhaap.<env>.orgs — skipping job template sync. ` +
+          'Add org names to enable catalog population (e.g., orgs: [Default]).',
+      );
+      return true;
+    }
```
abhikdps: we should throw this error in config.ts itself. the orgs array is already being fully parsed there. and i see this check being repeated both here and in the AAPEntityProvider as well. This was it will fail fast and eliminate duplication. Also this check makes more sense in the config.ts since it is the right layer as an empty orgs array is basically a misconfiguration and not a runtime condition.
The only place this would make sense in the provider file is to let the backend start even without an orgs being configured. if that is the intention then it makes sense here.

plugins/catalog-backend-module-rhaap/src/providers/config.ts
```diff
@@ -50,12 +51,30 @@ function readAapApiEntityConfig(
...
+    const seen = new Map<string, string>();
+    for (const org of organizations) {
+      const ns = formatNameSpace(org);
+      validateNamespace(ns, org);
+      const existing = seen.get(ns);
+      if (existing) {
+        throw new Error(
+          `Organization names "${existing}" and "${org}" both produce namespace "${ns}". Rename one to avoid collision.`,
+        );
+      }
```
abhikdps, line 88: will this ever be executed? since AAP itself does not allow multiple orgs with the same name. hence even after formatting the namespaces name, two different org/namespace names would never be the same. is there an edgecase?

app-config.yaml
```diff
@@ -49,7 +49,8 @@ backend:
     credentials: true
   database:
     client: better-sqlite3
-    connection: ':memory:'
+    connection:
+      directory: ./sqlite-data
```
abhikdps: if this change was made for testing purposes, please revert it.

abhikdps (changes requested):

plugins/backstage-rhaap-common/src/AAPClient/AAPClient.ts
audgirka: `fetchProfile` needs to distinguish 401 (expired token / revoked session) from other HTTP errors — it throws `AuthenticationError` on 401 which triggers the OAuth re-auth flow. `executeGetRequest` doesn't have this distinction — it throws a generic error on any non-ok response and specifically handles 403 but not 401. Using `executeGetRequest` here would lose the auth-specific error handling.

app-config.yaml
audgirka: reverted

plugins/catalog-backend-module-rhaap/src/providers/config.ts
audgirka: Yes — AAP prevents duplicate org *names* but `formatNameSpace()` normalizes names (lowercase, replace spaces/underscores with hyphens, strip special chars). Two different org names can produce the same namespace, e.g. `Platform Ops` and `Platform_Ops` both normalize to `platform-ops`. This check catches that at config parse time rather than at sync time with confusing entity collisions.

deploy/index.json
audgirka: done

plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.ts
audgirka: `orgs` is a required config field so this check was dead code. Removed the empty-orgs guard from both providers along with the unused `NO_ORG_CONFIG` / `NO_ORG_JOB_CONFIG` test fixtures. Namespace validation at sync start is kept since that serves a different purpose.

deploy/index.json
audgirka: Removed and added `deploy/` to `.gitignore`.

app-config.yaml
audgirka: Reverted.

deploy/index.json
audgirka: done

plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.ts
```diff
@@ -162,13 +173,27 @@ export class AAPJobTemplateProvider implements EntityProvider {
       }
 
       for (const { job, survey, instanceGroup } of aapJobTemplates) {
+        // Filter templates to configured orgs
```
ganeshrn, line 176: **[Medium]** Org-less job templates bypass tenant filter

When `job.summary_fields?.organization?.name` is undefined (system-level template not associated with any org), the `if (templateOrgName && ...)` check short-circuits to false, so the template is NOT filtered out. Then `ns` defaults to `'default'` at line 185. In a multi-org deployment, unscoped templates from unconfigured orgs bypass tenant isolation and appear in the catalog under the default namespace.

**Suggested fix** — skip templates with no org association in multi-org mode:
```ts
const templateOrgName = job.summary_fields?.organization?.name;
if (!templateOrgName && isMultiOrg) {
  continue; // Skip org-less templates in multi-org mode
}
if (
  templateOrgName &&
  !this.orgs.includes(templateOrgName.toLowerCase())
) {
  continue;
}
```

ganeshrn (commented):
## Code Review — 5 findings (2 high, 3 medium)

Reviewed against multi-org spec (PR #671) and ADRs 008/011. Two findings are inline; three are on unchanged lines in touched functions (included below).

---

### [High] `AAPEntityProvider.ts` ~line 223 — Dead-code tautological filter

The condition `if (user.is_orguser && !user.is_orguser)` is always `false` — `x && !x` is a tautological contradiction. The null-return branch is never reached, so no users are ever filtered out. Every user in `org.users` is included as an org member regardless of their `is_orguser` status.

**Suggested fix** — was this meant to filter non-org users?
```ts
if (!user.is_orguser) {
  return null;
}
```

---

### [High] `AAPEntityProvider.ts` ~line 308 — Undefined namespace produces malformed group ref

When a user's team is not found in any org's team list (the `!matched` branch), the code accesses `org.organization.namespace` — but the `Organization` interface has no `namespace` field. This pushes `undefined` into `userMembers`, which Backstage serializes as the string `'undefined'` in `spec.memberOf`, creating a dangling group reference that breaks catalog validation.

**Suggested fix:**
```ts
userMembers.push(`group:${getOrgNamespace(org.organization.name) ?? 'default'}/${org.organization.name}`);
```

---

### [Medium] `AAPLogoutHandler.tsx` ~line 38 — Swallowed signOut error creates split-brain auth

`originalSignOut()` errors are caught and silently swallowed, then the browser unconditionally redirects to AAP gateway logout. If `originalSignOut()` fails, the user's Backstage session cookies persist but the AAP session is cleared.

**Suggested fix:**
```ts
try {
  await originalSignOut();
} catch (err) {
  console.error('Failed to clear Backstage session during logout:', err);
}
```

plugins/catalog-backend-module-rhaap/src/providers/config.ts
```diff
@@ -50,12 +51,30 @@ function readAapApiEntityConfig(
...
+      .filter(o => o.length > 0);
   }
+
+  if (organizations.length > 1) {
+    const seen = new Map<string, string>();
+    for (const org of organizations) {
+      const ns = formatNameSpace(org);
+      validateNamespace(ns, org);
+      const existing = seen.get(ns);
+      if (existing) {
```
ganeshrn, line 84: **[Medium — spec violation]** Collision detection only runs for >1 org; no warning when `orgs` is absent or empty

When the `orgs` config key is absent, both `getString` and `getStringArray` throw, and `organizations` remains `[]`. The provider silently syncs zero entities — a data loss scenario. The spec requires: *"If omitted or empty, entity providers sync nothing and log a warning."*

Collision detection here correctly handles >1 org, but the empty case is missing.

**Suggested fix** — add before the collision check:
```ts
if (organizations.length === 0) {
  logger.warn(
    `No orgs configured in catalog.providers.rhaap.${envName} — entity sync disabled.`
  );
}
```

plugins/catalog-backend-module-rhaap/src/providers/AAPJobTemplateProvider.ts
audgirka: Validated on AAP 2.7: job templates always inherit `organization` from their `project` — `summary_fields.organization` is always present. AAP does not allow org-less job templates. The default namespace fallback (line 187) is a defensive guard that should never be reached in practice.

plugins/catalog-backend-module-rhaap/src/providers/config.ts
audgirka: Single-org mode returns `default` namespace unconditionally (getEffectiveNamespace returns 'default' when allOrgs.length <= 1), so `formatNameSpace` is never called and collision is not possible. The throw at 
config parse time (in config.ts line 68-76) is intentionally scoped to 'multi-org' where `formatNameSpace` conversion could produce duplicates.

Pavanjangle (approved):
Tested locally with rhdh-local on RHDH 1.10, both single-org and multi-org configurations.

**What was verified:**
- Multi-org catalog sync — Default, Engineering, SecOps orgs synced correctly, teams and users created in Backstage catalog
- Admin user entity page — team memberships spanning both default and non-default org namespaces
- Org group entity pages — all orgs load with correct `organization` type
- Template sync — job templates synced across orgs with correct `ansible.com/organization` annotation and namespace
- Template execution — launched job templates from different orgs successfully
- Non-admin user visibility — admin sees all templates, non-admin sees only permitted templates via AAP RBAC
- Single-org fallback — works as before when only Default org is configured
- Logout sync — AAPLogoutHandler clears session correctly across orgs

PR #412, opened 2026-06-23
fix: local MUI v4 UnregisterEntityDialog to fix broken BUI dialog

Pavanjangle:
## Summary

- Created a local MUI v4 `UnregisterEntityDialog` component inside `plugins/self-service/` to fix the broken backdrop and transparent background from the upstream MUI → BUI migration in `@backstage/plugin-catalog-react` v2.1.1
- Removed the previous CSS override approach (`packages/app/src/overrides.css`) which wouldn't ship with the dynamic plugin bundle
- Updated all 3 consumer components to import from the local component
- Added unit tests covering all dialog states

## Root cause

The upstream `UnregisterEntityDialog` was migrated from MUI v4 to `@backstage/ui` (BUI) in Backstage 1.48+. The BUI Dialog has:
- `--bui-bg-app` undefined → transparent dialog background
- `--bui-bg-neutral-2` nearly transparent → no visible backdrop overlay
- Dark mode selector mismatch: BUI uses `[data-theme="dark"]` but Backstage sets `[data-theme-mode="dark"]`

This is a known upstream bug tracked in [backstage/backstage#34325](https://github.com/backstage/backstage/issues/34325), fixed in Backstage 1.52.0. Our RHDH 1.10 is on an older version.

## Changes

| File | Change |
|------|--------|
| `plugins/self-service/src/components/UnregisterEntityDialog/` | New local MUI v4 component with same business logic (`catalogApiRef`) |
| `plugins/self-service/src/components/UnregisterEntityDialog/UnregisterEntityDialog.test.tsx` | 13 unit tests covering all dialog states |
| `CatalogItemDetails.tsx`, `CatalogContent.tsx`, `EEDetailsPage.tsx` | Updated imports to local component |
| `CatalogContent.test.tsx`, `EEDetailsPage.test.tsx` | Updated test mocks |
| `packages/app/src/overrides.css` | Removed |
| `packages/app/src/index.tsx` | Reverted CSS import |

## Scope

This fix covers the 3 places inside `plugins/self-service/` where `UnregisterEntityDialog` is used. The standard `/catalog` entity page uses the upstream dialog from `@backstage/plugin-catalog` which is outside our plugin scope — that will be fixed when RHDH picks up the upstream fix from Backstage 1.52.0.

## Testing

- Verified dialog renders correctly in both light and dark mode
- All 1911 tests passing (13 new + 1898 existing)
- TypeScript and lint clean

## Related

- Jira: https://redhat.atlassian.net/browse/AAP-79846
- Upstream bug: https://github.com/backstage/backstage/issues/34325
- Upstream fix: https://github.com/backstage/backstage/pull/34400
- Migration tracking: https://github.com/backstage/backstage/issues/31467

## Screenshot
Before Fix - 

Screenshot 2026-06-23 at 6 32 34 PM

After Fix -

Screenshot 2026-06-24 at 9 46 24 AM

## Summary by CodeRabbit

* **New Features**
  * Added an “Unregister Entity” confirmation dialog with clear options to unregister a location or delete an entity (including an “Advanced Options” delete flow).
  * Improved dialog behavior for special cases, showing appropriate loading/error states and guidance when an entity can’t be unregistered normally.
* **Bug Fixes**
  * Fixed entity unregister/delete actions to behave consistently across catalog-related views, including correct confirmation, success notifications, and error reporting.


NilashishC (commented):
## Review: CSS Override Approach

Thanks for the thorough root cause analysis — the identification of the `--bui-bg-neutral-2` transparency issue, the undefined `--bui-bg-app`, and the `data-theme-mode` vs `data-theme` selector mismatch is spot-on. However, I have significant concerns about this approach as a fix.

### 1. This fix does not reach production

`packages/app/src/overrides.css` lives in the local Backstage dev app. When the self-service plugin is built and exported as a dynamic plugin (`yarn export-dynamic`), only the contents of `plugins/self-service/` are bundled into the OCI image or tarball that ships to users. The `packages/app/` directory is excluded entirely — it is scaffolding for local development, not a deployable artifact.

The deployment chain is:

```
packages/app/src/overrides.css  ← lives here, never leaves this repo
        ✗ not included in
plugins/self-service/dist-dynamic/  ← what gets bundled
        → registry.redhat.io/.../automation-portal:2.2  ← OCI image
        → ansible-portal-chart deploys into
        → RHDH (production)  ← never sees overrides.css
```

End users on RHDH — whether deployed via the Helm chart, the Operator, or any other method — will still see the broken dialog. This fix only takes effect when a developer runs `yarn start` locally against this repo.

### 2. Global blast radius on all BUI Dialogs

The selectors `[class*='bui-DialogOverlay']` and `[class*='bui-Dialog']` target **every** BUI Dialog in the entire application, not just the `UnregisterEntityDialog`. As Backstage progressively migrates more components to BUI (this is their stated direction), these overrides will silently affect dialogs that were designed for the BUI styling system — potentially breaking components that render correctly under BUI's native theming.

### 3. Hardcoded color values diverge from RHDH theme

The dark mode overrides use hardcoded hex values (`#292929`, `#555`, `#9e9e9e`, `#9cc9ff`, `#7eb5f7`, `#e22b2b`, `#191919`, `#242424`) rather than being derived from the RHDH theme tokens. If RHDH changes its dark mode palette in a future release, these values will become stale and visually inconsistent with the rest of the UI. The same applies to the light mode `#fff` for `--bui-bg-app`.

### 4. `!important` overrides hide future upstream fixes

There are four `!important` declarations in the file. These will win over any future BUI CSS fixes that the Backstage team ships upstream. If Backstage fixes the `--bui-bg-app` resolution or the `data-theme-mode` selector mismatch in a future release (which they likely will, since it affects all BUI consumers), our overrides will mask those improvements and we'd need to remember to remove them.

### 5. Only addresses visual issues, not behavioral differences

The underlying problem is that the upstream `UnregisterEntityDialog` in `@backstage/plugin-catalog-react` v2.0.0+ was rewritten from MUI v4 to `@backstage/ui` (BUI) — commit `0e9578d93e65`, shipped in Backstage v1.48.0. This wasn't just a reskin; it changed:

- **Interaction pattern**: The "Advanced Options" toggle button + `<Divider>` was replaced with a `<Accordion>` / `<AccordionTrigger>` / `<AccordionPanel>` — a fundamentally different UX
- **Dialog chrome**: `<DialogHeader>` now renders a built-in close X button (via `<RiCloseLine />` from Remix icons) that the old `<DialogTitle>` didn't have
- **Button semantics**: MUI v4 `onClick` / `disabled` replaced with React Aria's `onPress` / `isDisabled` — different keyboard interaction behavior
- **Focus management**: React Aria's `Modal` focus trap behaves differently from MUI v4's Dialog focus trap
- **Footer layout**: `<DialogFooter>` is right-aligned with a `border-top` separator; the old `<DialogActions>` was inline with no separator

The CSS overrides fix the backdrop and background colors, but the dialog still looks and behaves like a BUI component surrounded by MUI v4 UI — the accordion, the close button, the button styling, and the footer layout all remain visually inconsistent with the rest of the self-service plugin.

### Recommended approach

The fix should live inside `plugins/self-service/` so it ships with the dynamic plugin bundle. A local MUI v4 reimplementation of the `UnregisterEntityDialog` component — using the public `catalogApiRef` API from `@backstage/plugin-catalog-react` for the business logic, but rendering with MUI v4 `Dialog`, `Button`, `Alert`, etc. — would:

- Ship to production (bundled in the plugin)
- Affect only the 3 pages that use it (no global blast radius)
- Restore full visual and behavioral parity with the pre-migration UX
- Require no `!important` overrides or hardcoded color values
- Be self-contained and maintainable until a broader BUI migration is undertaken

NilashishC (changes requested):
Are you sure this fix will work with the self-service plugin built and exported as a dynamic plugin?

Additionally, this was merely a visual fix. I'm fairly certain that the Backstage's breaking change related to this also introduced behavioral changes as well for this dialog. Those do not seem to be fixed here.

Since the 1.10 upgrade for ansible-backstage-plugins is scheduled to be a part of a z-stream release, there MUST NOT be any breaking changes (behavioral + UI/UX), compared to the last 2.2.x release that goes out before it.

I'm a -1 for this approach anyway since this has a global blast radius. The selectors `[class*='bui-DialogOverlay']` and `[class*='bui-Dialog']` match every BUI Dialog in the application, not just the `UnregisterEntityDialog`. These overrides would apply to all of them, potentially breaking dialogs that were designed to work with the BUI token system.

Pavanjangle:
> Are you sure this fix will work with the self-service plugin built and exported as a dynamic plugin?
> 
> Additionally, this was merely a visual fix. I'm fairly certain that the Backstage's breaking change related to this also introduced behavioral changes as well for this dialog. Those do not seem to be fixed here.
> 
> Since the 1.10 upgrade for ansible-backstage-plugins is scheduled to be a part of a z-stream release, there MUST NOT be any breaking changes (behavioral + UI/UX), compared to the last 2.2.x release that goes out before it.
> 
> I'm a -1 for this approach anyway since this has a global blast radius. The selectors `[class*='bui-DialogOverlay']` and `[class*='bui-Dialog']` match every BUI Dialog in the application, not just the `UnregisterEntityDialog`. These overrides would apply to all of them, potentially breaking dialogs that were designed to work with the BUI token system.

@NilashishC thanks for the review. agreed the CSS override approach won't ship with the dynamic plugin bundle and has global blast radius issues. creating a local MUI v4 reimplementation inside `plugins/self service/` using `catalogApiRef` for business logic. will update the PR.

NilashishC:
## Review notes — second iteration

The local MUI v4 reimplementation directly addresses all the concerns raised in the first review. A few observations from a closer look at the updated code:

**Optional improvements**

- `UnregisterEntityDialogProps` and `DialogState` are the only public exports from the new module. Adding minimal JSDoc to both types would address the CodeRabbit docstring coverage warning and improve DX for future consumers.

- The bootstrap state's "Advanced Options → Delete Entity" happy path (successful delete from within `BootstrapBody`) isn't covered by a unit test. The existing bootstrap test only verifies the info alert renders. Adding a test for that branch would close the Codecov gap in `UnregisterEntityDialog.tsx` (currently at 91.89% patch coverage).

**Informational**

- As already documented in the PR description, the standard `/catalog` entity page still uses the broken upstream BUI dialog from `@backstage/plugin-catalog`. This is acceptable for the z-stream scope — it will self-heal when RHDH picks up the upstream fix from Backstage 1.52.0.

- The CodeRabbit `getLocationByRef` null-guard issue was correctly fixed in `3a102eb`. All CI checks (SonarCloud, CodeQL, pre-commit, tests on Node 20 and 24) pass on the current head.

The implementation is correctly scoped to `plugins/self-service/`, uses MUI v4 throughout (consistent with the rest of the plugin), and will ship with the dynamic plugin bundle.

sshedmake (approved):
LGTM.

Verified the fix, via rhdh-local, on rhdh-1.10 with Node 24. Here's the screenshot.

Screenshot From 2026-06-26 23-35-29

Pavanjangle:
@NilashishC Based on your second-iteration feedback, i incorporated the suggested changes. Can you re-review and approve (if it looks good).

Pavanjangle:
Tested with Jenkins CI too. Build #558 and Logs - [#558.txt](https://github.com/user-attachments/files/29402111/558.txt)

PR #433, opened 2026-07-02
feat(e2e): add GitLab EE build template execution and detail view tests

abhikdps:
## Description

Adds Playwright E2E tests for the GitLab-based Execution Environment build template workflow and the EE catalog detail view.

### Changes

- This tests gets skipped in CI due to know limitations with Cloudflare Turnstile challenge on GitLab. Can be run manually.
- **`ee02b-gitlab-template-execution.spec.ts`**: End-to-end test that exercises the full GitLab EE build template: filling out the scaffolder form (name, EE image, base image, collections, Galaxy credentials), handling GitLab OAuth login (popup and redirect flows), and verifying task completion
- **`ee03-detail-view.spec.ts`**: Tests for the EE catalog and detail view: validates the Catalog tab (empty state or table), sidebar filters, and detail page navigation
- **`utils/auth.ts`**: Shared authentication utilities for RH AAP sign-in
- **`.env.example`**: Adds `GL_USER_ID`, `GL_USER_PASS`, and `GL_AUTHENTICATOR_SECRET` placeholders for GitLab credentials

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Testing

- Tests run via `yarn test:e2e` with Playwright
- Requires GitLab credentials (`GL_USER_ID`, `GL_USER_PASS`) and a running RHDH instance with GitLab OAuth configured
- Tests handle both popup and redirect OAuth flows for GitLab authentication

## Checklist

- [x] Tests pass locally
- [x] Documentation updated (`.env.example`)

## Summary by CodeRabbit

* **New Features**
  * Added GitLab end-to-end coverage for self-service template execution, including OAuth “login required” handling and Cloudflare-style security verification.
  * Added GitLab-published execution environment build test coverage from the catalog/detail view.
  * Expanded E2E authentication helpers to support GitLab (including optional two-factor).

* **Bug Fixes**
  * Updated the default execution environment template source to v2.0.1.
  * Improved E2E build reliability with randomized suffix-based image naming.

* **Documentation**
  * Updated the E2E environment example with GitLab credential and optional 2FA/namespace settings.


e2e-tests/playwright/tests/self-service/ee02-template-execution.spec.ts
```diff
@@ -231,7 +231,7 @@ async function handleGitHubLoginOnPage(page: Page): Promise<void> {
 
 const EE_TEMPLATE_URL =
   process.env.EE_IMPORT_REPO_URL ||
-  'https://github.com/ansible/ansible-rhdh-templates/blob/v1.0.2/templates/ee-start-from-scratch.yaml';
+  'https://github.com/ansible/ansible-rhdh-templates/blob/main/templates/ee-start-from-scratch.yaml';
```
NilashishC: **[Medium]** This changes the default template URL from the pinned `v1.0.2` tag to `main`. If the template repo's `main` branch introduces a breaking schema change, both the existing GitHub test (ee02) and the new GitLab test (ee02b) will break.

Was this intentional? If so, a brief comment explaining why `main` is preferred over a stable tag would help future maintainers. If not, consider keeping the pinned tag for test stability.

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
```diff
@@ -0,0 +1,1023 @@
...
+  }
+
+  console.log('[EE GitLab Test] Clicking Log in button...');
+  console.log('[EE GitLab Test] URL before click:', page.url());
+
+  const context = page.context();
+
+  const [popup, navigationOrNull] = await Promise.all([
+    context.waitForEvent('page', { timeout: 3000 }).catch(() => null),
+    page.waitForNavigation({ timeout: 3000 }).catch(() => null),
```
NilashishC: **[Low]** `page.waitForNavigation()` was deprecated in Playwright 1.27 in favour of `page.waitForURL()`. I see the same pattern in the existing ee02 GitHub test, so this is consistent — but since you're writing new code, it'd be nice to use the modern API.

The subsequent code already handles URL detection (checking `page.url()` hostname), so you could simplify this to just fire the click and rely on the explicit URL checks below.

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
```diff
@@ -0,0 +1,1023 @@
...
+        .first()
+        .fill('execution environment');
+
+      const oauthResult = await handleGitLabOAuthDialog(page);
+
+      if (oauthResult === 'failed') {
+        console.log(
+          '[EE GitLab Test] GitLab OAuth failed — skipping Git publish flow, deferring to non-Git run',
+        );
+      } else if (oauthResult === 'redirected') {
```
NilashishC: **[Medium — maintainability]** The wizard-fill logic from here through line ~896 (provider re-selection, EE Definition Name/Description, namespace selection, repo name, Build EE toggle) is repeated three times across the `redirected` branch, the `else` branch (line 828), and the second run (line 898).

Consider extracting a helper like:

```ts
async function fillEeWizardFields(page: Page, opts: {
  withGitPublish: boolean;
  selectNamespace?: boolean;
}) { ... }
```

This way a wizard UI change only needs updating in one place. The existing GitHub test (ee02) has a similar pattern, so this isn't unique to this PR — but since this is ~200 lines of near-verbatim duplication, it's worth cleaning up.

e2e-tests/playwright/tests/self-service/ee03-detail-view.spec.ts
```diff
@@ -435,6 +435,106 @@ test.describe('Execution Environment Catalog and Detail View Tests', () => {
...
+      const toast = page.locator(
+        '[role="alert"], .MuiSnackbar-root, [class*="toast" i], [class*="notification" i]',
+      );
+      await expect(toast.first()).toBeVisible({ timeout: 15000 });
+
+      const toastText = await toast.first().innerText();
+      expect(toastText).toContain('Build triggered');
+
+      // Verify the toast contains a GitLab pipeline URL
+      const toastLink = toast.first().locator('a[href*="gitlab"]');
```
NilashishC: **[Low — functional correctness]** The test title says "shows pipeline URL" but this check is best-effort: when the GitLab link isn't found in the toast, it logs and passes silently. The test can never actually fail on the pipeline URL assertion.

Two options:
1. **Make it strict** — assert the link is present: `await expect(toastLink.first()).toBeVisible({ timeout: 5000 });`
2. **Keep it soft but rename** — change the test title to something like "Build on GitLab-published EE triggers build (pipeline URL optional)" so the intent is clear.

Either is fine — the current state just sends mixed signals about what the test guarantees.

NilashishC (commented):
Good work adding GitLab E2E coverage — the test structure is consistent with the existing GitHub tests and handles the OAuth flows (popup, redirect, Turnstile) thoroughly.

A few things to consider before merge — see inline comments for details:

1. **Template URL unpin** (`v1.0.2` → `main`) — intentional? This affects the existing GitHub test too.
2. **Wizard-fill duplication** — ~200 lines repeated 3× across OAuth branches. A helper function would reduce maintenance burden.
3. **Pipeline URL assertion** — currently best-effort but test title implies strict verification.
4. **Deprecated API** — `waitForNavigation()` could be replaced with `waitForURL()` since you're writing new code.

None of these are blockers on their own — the tests will work as-is and provide valuable coverage. The template URL question is the one I'd most like clarified.

e2e-tests/playwright/tests/self-service/ee02-template-execution.spec.ts
abhikdps: changed to v2.0.1

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
abhikdps: Fixed. Removed the `waitForNavigation`/`waitForURL` from `Promise.all` entirely and simplified to `Promise.race` for popup detection. The downstream code already handles URL detection via explicit hostname checks, so the navigation wait was redundant. This addresses the deprecation without introducing the semantic regression that `waitForURL(/.*/)` caused (it returns void instead of Response, breaking the truthiness check).

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
abhikdps: Fixed. Extracted five helper functions: `selectGitLabProvider`, `fillWizardCommonFields`, `fillPublishFields`, `findAndStartTemplate`, and `waitForTaskCompletion`. This eliminated ~300 lines of duplicated wizard-fill logic. A wizard UI change now only needs updating in one place.

e2e-tests/playwright/tests/self-service/ee03-detail-view.spec.ts
abhikdps: Fixed. Made the assertion strict with `await expect(toastLink.first()).toBeVisible({ timeout: 5000 })`. The test now fails if the GitLab pipeline URL is missing from the toast.

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
```diff
@@ -0,0 +1,883 @@
...
+
+  console.log('[EE GitLab Test] Clicking Log in button...');
+  console.log('[EE GitLab Test] URL before click:', page.url());
+
+  const context = page.context();
+
+  const popup = await Promise.race([
+    context.waitForEvent('page', { timeout: 3000 }).catch(() => null),
+    logInBtn.click().then(() => null),
+  ]);
```
NilashishC, line 38: **Popup detection race condition** — `Promise.race` here means that if `click()` resolves first (which it normally does, since Playwright resolves click after dispatching the action), the popup listener is abandoned immediately. A popup that opens slightly later (common with OAuth redirects) would be missed, forcing the test into the redirect fallback path unnecessarily.

The GitHub variant (`ee02-template-execution.spec.ts:43-47`) uses `Promise.all` which keeps both listeners alive for the full 3s timeout:

```ts
const [popup, navigationOrNull] = await Promise.all([
  context.waitForEvent("page", { timeout: 3000 }).catch(() => null),
  page.waitForNavigation({ timeout: 3000 }).catch(() => null),
  logInBtn.click(),
]);
```

Consider switching to `Promise.all` (or kick off the `waitForEvent` before calling `click()`) so that popup detection has the full timeout window regardless of when `click()` resolves.

Non-blocking since the redirect path is a valid fallback, but it makes the popup path effectively unreachable in practice.

e2e-tests/playwright/tests/self-service/ee03-detail-view.spec.ts
```diff
@@ -435,6 +438,143 @@ test.describe('Execution Environment Catalog and Detail View Tests', () => {
...
+      await imageNameInput.first().fill(`ee-test/ee-gl-${suffix}`);
+      await page.waitForTimeout(500);
+    }
+
+    // Click Build button
+    const buildButton = modal
+      .getByRole('button', { name: /build/i })
+      .or(modal.locator('button').filter({ hasText: /build/i }));
+
+    if ((await buildButton.count()) > 0) {
```
NilashishC: **Build button guard allows silent pass** — once the modal is open (line 535), the entire build trigger + GitLab pipeline URL assertion is gated behind this `count > 0` check. If the Build button does not render, the test passes silently without verifying anything.

Since earlier guards already `test.skip` when the EE row or Build menu item is absent, reaching the modal means the build flow should proceed. Consider adding an `else` branch that throws an error like: `Build button not found in modal — GitLab build flow cannot be verified`.

(CodeRabbit flagged this too — just confirming it is still present in the latest commit.)

e2e-tests/playwright/utils/auth.ts
```diff
@@ -277,6 +277,121 @@ export async function loginAAPWithSession(page: Page) {
...
+  if (!page.url().includes('/ansible')) {
+    await page.goto('/ansible', { waitUntil: 'domcontentloaded' });
+  }
+
+  if (await signInGate.isVisible({ timeout: 3000 }).catch(() => false)) {
+    throw new Error(
+      'Still on portal sign-in picker after GitLab flow; check OAuth app / credentials.',
+    );
+  }
+}
```
NilashishC, line 582: Nit: `signInRHDHWithGitLab` (lines 322-393) is structurally identical to `signInRHDHWithGitHub` (lines 450-523) — they differ only in the login function called, the provider filter text, and the error message. A parameterized helper like `signInRHDHWithProvider(page, { login, providerName })` would halve the code and keep both paths in sync.

Not blocking — fine as a follow-up.

NilashishC (commented):
Good work on the GitLab E2E coverage — the test structure is consistent with the existing GitHub tests and the OAuth + Turnstile handling is thorough. Nice job extracting the wizard helpers (`fillWizardCommonFields`, `selectGitLabProvider`, `fillPublishFields`, `findAndStartTemplate`) to reduce duplication.

A couple of things to consider — see inline comments:

1. **Popup detection race** (`Promise.race` vs `Promise.all`) in `handleGitLabOAuthDialog` — the popup listener gets abandoned if `click()` resolves first, making the popup path effectively unreachable.
2. **Silent pass on Build button** in the GitLab build test — an `else throw` would prevent the test from passing without exercising its core assertions.

Neither is strictly blocking — the tests work as-is and provide valuable coverage. The popup issue has a working fallback (redirect path), and the build button guard is unlikely to trigger in practice since earlier checks already skip when prerequisites are missing.

e2e-tests/playwright/tests/self-service/ee02b-gitlab-template-execution.spec.ts
abhikdps: Fixed. Switched back to `Promise.all` so the popup listener has the full 3s timeout window regardless of when `click()` resolves. This matches the pattern in the GitHub test.

e2e-tests/playwright/tests/self-service/ee03-detail-view.spec.ts
abhikdps: Fixed. Flipped to `throw new Error('Build button not found in modal')` when the button is absent.

e2e-tests/playwright/utils/auth.ts
abhikdps: extracted `signInRHDHWithProvider(page, login, providerName)` as a shared helper. Both `signInRHDHWithGitLab` and `signInRHDHWithGitHub` are now one-liners delegating to it. Also moved the GitLab OAuth page handler (Turnstile + credentials + TOTP + Authorize) from `ee02b` into `auth.ts` as exported functions so `ee03`'s build flow can reuse them.

NilashishC (approved):
All prior feedback addressed — nice work on the refactoring. The helper extraction (`fillWizardCommonFields`, `selectGitLabProvider`, `fillPublishFields`, `findAndStartTemplate`) and the consolidated `signInRHDHWithProvider` in auth.ts are real improvements. Template URL pinned to `v2.0.1`, deprecated API removed, assertions are explicit. LGTM.

PR #466, opened 2026-07-09
[ANSTRAT-912] feat: per-template RBAC visibility filtering

audgirka:
## Description

Per-template visibility filtering at the database level using a custom Backstage permission rule. Users see only AAP job templates they have execute permission on. Custom/SCM templates remain visible to all.

## Related Issues

[ANSTRAT-912](https://redhat.atlassian.net/browse/ANSTRAT-912)
Depends on #408 (multi-org namespace scoping)
ADR-011: Per-Template RBAC Filtering (PR ansible/ansible-rhdh-plugins#671)

## Type of Change

- [x] New feature

## Testing

31 new unit tests (all 3562 tests pass):
- ExecutePermissionStore (10)
- HAS_EXECUTE_PERMISSION rule (7)
- AAPJobTemplateProvider store population (2)
- AAPClient.getJobTemplateExecuteMap (3)
- AAPRBACProvider (9)

Validated on OCP dev cluster with 2 replicas (HA):
- portal-user sees 2 AAP templates (1 direct + 1 team)
- network-user sees 3 AAP templates (direct execute)
- admin sees all 29 (superuser bypass)
- Custom templates visible to all

## Checklist

- [x] Tests pass locally
- [x] Documentation updated


plugins/permission-backend-module-rhaap/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@ansible/backstage-plugin-permission-backend-module-rhaap",
+  "description": "RBAC provider for AAP multi-org template visibility.",
+  "version": "2.2.3",
```
abhikdps, line 4:
```suggestion
  "version": "2.2.4",
```

abhikdps (changes requested):
The only change needed is the new plugin version. rest looks good!

plugins/permission-backend-module-rhaap/package.json
audgirka: @abhikdps this would eventually be changed to the next release version

craig-br:
**Breaking change review — per-template RBAC + default roles**

Thanks for the implementation. From a customer upgrade perspective, a few items need addressing before merge:

**1. Per-template visibility (confirmed breaking, intended)**  
Non-admin users will only see AAP job templates where they hold JobTemplate Execute/Admin in AAP. This is the right security posture but must be called out explicitly in release notes — single-org customers are affected too (`659343d`), not only multi-org.

**2. Default portal personas — admin / executor / auditor (needs ADR + migration guide)**

PR #466 auto-creates only `role:default/aap-user` (executor-like). Product intent references auditor and admin personas (APME-PROD-03, migration guide `portal-viewer`/`deployer` examples). Please clarify the GA persona model:

- **Admin:** `aap-admins` / superuser — looks complete.
- **Executor:** `aap-user` + `HAS_EXECUTE_PERMISSION` — maps to AAP JT Execute/Admin via **existing teams**. Confirm: **customers do NOT need new portal teams per org** for this PR.
- **Auditor:** **Not implemented.** Execute-only filtering means AAP org members/auditors without JT Execute see zero AAP templates. Is auditor in or out of GA scope?

Open questions:
- Rename `aap-user` → `aap-executor` for clarity?
- Deprecate `portal-users` in `default-rbac-policy.csv`?
- Ship three-role reference CSV before GA?
- How should AAP System Auditor map to portal read-only access?

Suggest ADR-012 or ADR-011 appendix with persona ↔ AAP role matrix and explicit "no portal-specific teams required" guidance.

**3. Missing upgrade path for new dynamic plugin**  
Please add documentation (README on `permission-backend-module-rhaap`, CHANGELOG, and/or `dist-dynamic/README.md`) covering:
- RHDH `dynamic-plugins.yaml` backend entry for `@ansible/backstage-plugin-permission-backend-module-rhaap`
- OCP/Helm and bootc appliance upgrade steps
- Failure mode if catalog module is upgraded without the permission module

**4. AAP 2.5+ gateway API requirement**  
`getJobTemplateExecuteMap()` depends on `api/gateway/v1/role_*_assignments`. Please document minimum AAP version (2.5+) and required token scopes in plugin README / config docs.

**5. Startup window**  
ADR-011 notes ~30–60s where execute store is empty and all AAP templates are hidden post-restart. Worth a release-note callout for rolling upgrades.

**6. ANSTRAT-1237 alignment — template-source annotation convention**

Add `metadata.annotations['ansible.com/template-source']: aap` in `dynamicJobTemplate.ts` alongside `spec.type: aap-template`. Annotation = platform only; spec.type = kind. Amend ANSTRAT-1237 from `aap-template-provider` → `aap`.

**7. Extend ADR-009 with reserved taxonomy types**

Replace `job-template` with `aap-template` in ADR-009 and #466. Reserve future types now so we do not overload `aap-template` when workflow/orchestrator catalog entries land:
- `aap-workflow` ([ANSTRAT-1651](https://redhat.atlassian.net/browse/ANSTRAT-1651) — AAP Workflow JTs)
- `orchestrator-workflow` ([ANSTRAT-1938](https://redhat.atlassian.net/browse/ANSTRAT-1938) — Orchestrator starter templates)


**8. Minimal config / sane defaults for greenfield install**

A default portal install must ship with workable defaults — customers should not need to discover optional-looking config before the catalog syncs. In practice:

- Fresh install (bootc, Helm, reference `app-config`) should include `orgs: [Default]` (or equivalent) when connecting to a standard single-org AAP instance whose default organization is named **Default**.
- Empty/omitted `orgs` stopping all sync is correct for **misconfiguration detection**, but greenfield installs must not land in an empty-catalog state by default.
- ADR/spec should state explicitly: **portal should require minimal configuration to install and run, with sane defaults** — explicit org list for upgrades/multi-org, but a single-org Default org should work out of the box.

Please confirm reference configs, bootc, and Helm values ship `orgs: Default` and document this as the expected greenfield baseline.

plugins/permission-backend-module-rhaap/src/module.ts
```diff
@@ -0,0 +1,175 @@
...
+    permissionMapping: [{ name: 'catalog.entity.read', action: 'read' }],
+  };
+}
+
+/** @internal Exported for testing only */
+export function formatNamespace(name: string): string {
+  return name
+    .toLowerCase()
+    .replace(/[^a-z0-9-]/g, '-')
+    .replace(/-+/g, '-')
```
ganeshrn, line 68: **[Critical — cross-PR bug]** `formatNamespace` diverges from catalog's `formatNameSpace` — breaks RBAC for orgs with dots or underscores

PR #466 introduces a separate `formatNamespace()` here that converts dots to dashes. But PR #408's `formatNameSpace()` in `helpers.ts` strips dots entirely. For an org named `"Acme.Corp"`:
- Permission module (this PR): `acme-corp` (dots → dashes)
- Catalog module (PR #408): `acmecorp` (dots stripped)

The RBAC provider assigns `group:acme-corp/acme-corp → role:default/aap-user`, but catalog entities live in namespace `acmecorp`. The role membership never matches — all templates become invisible for that org despite users having execute permissions.

**Suggested fix** — import and reuse the canonical function from helpers.ts instead of duplicating:
```ts
import { formatNameSpace } from '../helpers';
// Remove the local formatNamespace function entirely
```

plugins/catalog-backend-module-rhaap/src/permissions/executePermissionStore.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  hasExecutePermission(username: string, templateId: string): boolean {
+    return this.userToTemplates.get(username)?.has(templateId) ?? false;
+  }
+
+  get size(): number {
+    return this.userToTemplates.size;
+  }
+}
```
ganeshrn, line 27: **[High]** Module-level singleton store is overwritten by last provider to refresh in multi-env deployments

The `executePermissionStore` singleton is shared across all `AAPJobTemplateProvider` instances. In multi-environment deployments (`rhaap.prod` + `rhaap.staging`), both providers import this same instance. When provider `staging` calls `update()`, it builds a completely new map and replaces `this.userToTemplates` entirely — `prod`'s permissions disappear until `prod`'s next refresh (up to 40 minutes later).

**Suggested fix** — make the store per-environment by merging rather than replacing:
```ts
update(newMappings: Map<string, string[]>) {
  for (const [key, val] of newMappings) {
    this.userToTemplates.set(key, val);
  }
}
```
Or create separate store instances per environment in the module registration.

plugins/catalog-backend-module-rhaap/src/permissions/rules.ts
```diff
@@ -0,0 +1,40 @@
...
+  name: 'HAS_EXECUTE_PERMISSION',
+  description:
+    'Allow job template entities where the user has AAP execute permission',
+  resourceRef: catalogEntityPermissionResourceRef,
+  paramsSchema,
+  apply: (
+    resource: { metadata: { aapJobTemplateId?: number | string } },
+    { userEntityRef }: { userEntityRef: string },
+  ) => {
+    const templateId = String(resource.metadata.aapJobTemplateId ?? '');
```
ganeshrn, line 24: **[High]** Persistent API failure keeps the store empty forever, permanently hiding all AAP templates

If `getJobTemplateExecuteMap()` fails persistently (expired AAP token, network issues), the store stays empty and all AAP templates remain invisible. There's only a warn-level log with no retry escalation, no error counter, no health check. Operators have no signal that RBAC filtering is broken.

**Suggested fix** — add failure tracking and escalate logging:
```ts
if (this.consecutiveFailures >= 3) {
  logger.error(
    `ExecutePermissionStore failed ${this.consecutiveFailures} consecutive refreshes — ` +
    `RBAC filtering may be stale. Check AAP connectivity and token validity.`
  );
}
```
Also consider a circuit-breaker: after N failures, fall back to permissive mode (ALLOW all) rather than silently hiding all templates.

plugins/permission-backend-module-rhaap/src/module.ts
```diff
@@ -0,0 +1,175 @@
...
+];
+
+function buildConditionalPolicy(): RoleConditionalPolicyDecision<PermissionInfo> {
+  const decision: ConditionalPolicyDecision = {
+    result: 'CONDITIONAL' as const,
+    pluginId: 'catalog',
+    resourceType: 'catalog-entity',
+    conditions: {
+      anyOf: [
+        {
```
ganeshrn, line 39: **[Medium — spec gap]** Username extraction assumes Backstage user entity name exactly matches AAP username

`userEntityRef.split('/').pop()` extracts the username for AAP permission lookups. If Backstage uses email-based entity refs (`user:default/john.doe@example.com`) but AAP stores `john.doe` (sAMAccountName), the store lookup fails — the user sees zero templates with no error logged.

This is common in LDAP/SAML deployments where Backstage uses email as the entity name while AAP uses sAMAccountName.

**Suggested fix** — at minimum, document this assumption. Ideally, add a configurable username transform or strip the email domain:
```ts
const username = userEntityRef.split('/').pop()?.split('@')[0];
```

plugins/catalog-backend-module-rhaap/src/permissions/rules.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!templateId) return true;
+    const username = userEntityRef.split('/').pop() ?? '';
+    return executePermissionStore.hasExecutePermission(username, templateId);
+  },
+  toQuery: ({ userEntityRef }: { userEntityRef: string }) => {
+    const username = userEntityRef.split('/').pop() ?? '';
```
ganeshrn, line 30: **[Medium]** `toQuery()` returns string template IDs but entity metadata stores `aapJobTemplateId` as a number — type mismatch risk

`toQuery()` returns `{key: 'metadata.aapJobTemplateId', values: ['11']}` (strings). But `dynamicJobTemplate.ts` stores `metadata.aapJobTemplateId = job.id` (number). The `apply()` method correctly uses `String()` conversion, but `toQuery()` is used for efficient batch filtering in the catalog — if the catalog uses strict equality or SQL type-sensitive comparison, the filter never matches and all templates become invisible.

**Suggested fix** — ensure consistent types at the source:
```ts
// In dynamicJobTemplate.ts
aapJobTemplateId: String(job.id),
```

ganeshrn (commented):
## Code Review — 5 findings (1 critical, 2 high, 2 medium)

Review of per-template RBAC visibility filtering — checked against ADR-011, spec FR-023/FR-025, and cross-referenced with PR #408's namespace handling.

**Critical:** `formatNamespace` diverges from PR #408's `formatNameSpace` — orgs with dots/underscores get different namespaces, breaking RBAC bindings
**High:** Singleton store overwritten in multi-env, persistent API failure hides all templates forever
**Medium:** Username extraction mismatch (email vs sAMAccountName), toQuery string/number type mismatch

plugins/permission-backend-module-rhaap/src/module.ts
audgirka: Fixed — added dot to the regex (`[_.\s]+` → dots now convert to dashes) for consistency with PR #408's `formatNameSpace`. Both PRs now produce identical namespace conversions.

plugins/catalog-backend-module-rhaap/src/permissions/executePermissionStore.ts
audgirka: The singleton is intentional — one store per backend process, updated atomically by `update()` which rebuilds the full map. Each `update()` replaces the entire map with a fresh snapshot from `getJobTemplateExecuteMap()`, which queries all configured orgs in a single pass. There's no partial state or race condition. Multiple provider configs would call `update()` sequentially during their sync cycles, and the last update reflects the latest AAP state.

plugins/catalog-backend-module-rhaap/src/permissions/rules.ts
audgirka: Investigated: `getJobTemplateExecuteMap()` calls `api/gateway/v1/role_user_assignments/` and `api/gateway/v1/role_team_assignments/` — both gateway endpoints. OAuth login also goes through the gateway (`/o/authorize/`, `/o/token/`). If the gateway is down, users can't authenticate — so nobody is logged in to query the catalog. The scenario where templates sync but permissions don't can't happen in practice since all APIs (org sync, user sync, template sync, permissions sync, and auth) go through the same gateway. On subsequent sync failures, the previous store data is retained since `update()` is only called on success (inside the try block).

plugins/permission-backend-module-rhaap/src/module.ts
audgirka: Validated: usernames are used raw throughout all layers — `entityParser` sets `name: user.username` without normalization (line 161), the execute store keys use raw AAP usernames from `getJobTemplateExecuteMap()`, and the RBAC lookup extracts the raw name from entityRef via `split('/').pop()`. Backstage's `isValidEntityName` allows `[a-zA-Z0-9]` separated by `[-_.]` (dots, dashes, underscores all valid in `metadata.name`). This is different from `metadata.namespace` which only allows `[a-z0-9][-]`. Since usernames are entity names (not namespaces), dots in usernames like `john.doe` are preserved consistently. The `formatNameSpace` dots-to-dashes conversion is only applied to org names for namespace conversion, never to usernames.

plugins/catalog-backend-module-rhaap/src/permissions/rules.ts
audgirka: `apply()` uses `String(resource.metadata.aapJobTemplateId)` which converts any type to string for comparison. `toQuery()` returns string values in the `values` array. Backstage catalog serializes all metadata values as strings after ingestion — the `aapJobTemplateId` annotation is set as `String(template.id)` in `dynamicJobTemplate.ts` (line 527). Types align at all comparison points.

audgirka:
Thanks @craig-br  — addressed immediate items:

**#1 Per-template visibility (breaking)**: Agreed — called out in migration guide (PR #671) with dedicated "Single-org visibility changes" section. Release notes callout needed.

**#2 Default personas**: Role renamed from `aap-user` to `aap-normal-user` aligning with AAP's three roles (admin, normal user, auditor). Auditor role scoped as follow-up — straightforward addition, discussed in refinement. No new portal-specific teams required — filtering uses existing AAP team and user role assignments.

**#3 Plugin upgrade path**: Install steps documented in migration guide (PR #671) for Helm (`dynamic-plugins.yaml` entry) and bootc (pre-baked, no action). Failure mode documented — without the permission module, templates visible to all (pre-upgrade behavior).

**#4 AAP 2.5+ gateway API**: Portal already requires AAP 2.5+. Documented in migration guide. No new version requirement.

**#5 Startup window**: Documented in migration guide with a Note in the overview section.

**Follow-up items (not blocking Early Access):**

- **#6/#7 Template taxonomy rename + annotation**: Recommend separate ADR-012 for full taxonomy including `aap-workflow` (ANSTRAT-1651) and `orchestrator-workflow` (ANSTRAT-1938). Renaming `job-template` → `aap-template` mid-PR would affect PR #408 as well.
- **#8 Minimal config / sane defaults**: Will discuss in next refinement — whether greenfield installs should ship `orgs: [Default]` by default.

PR #392, opened 2026-06-10
feat: Bump Backstage to 1.49.4 & add node.js 24 support

sshedmake:
## Summary
- Add Node.js 24 to supported engines (`22 || 24`) and CI matrix
- Upgrade `better-sqlite3` from `^9` to `^12` for Node 24 native module compatibility
- Remove `isolated-vm@5.0.4` resolution pin (no longer needed)
- Bump Backstage from 1.45.2 to 1.49.4 via `backstage-cli versions:bump --release 1.49.4`
- Fix all breaking changes from the Backstage bump:
  - Remove deprecated `variant="gridItem"` props from catalog cards
  - Update `createTemplateAction` schema API to use new `z =>` callback style
  - Replace unsafe `as any` casts with proper type annotations (`OAuthAuthenticator<unknown, GithubProfile>` for githubAuthenticator, pre-typed `ZodType` variable for EE definition schema)
  - Add required test dependencies for Backstage CLI 0.36.0 (`jest@^29`, `jest-cli@^29`, `jest-environment-jsdom@^29`, `@backstage/cli-defaults`, `@backstage/frontend-test-utils`)
- Remove dead schema exports: delete `helpers/schemas.ts` (10 unused exports), remove `aapApiRecordOutputSchema` and `launchJobTemplateValuesLooseSchema` from `rhaapActionSchemas.ts`, unexport `organizationSchema` (internal-only)

## Test plan
- [x] `yarn tsc` passes with zero errors
- [x] `yarn lint` passes clean
- [x] `yarn test` passes (all 16 test suites)
- [x] All 5 dynamic plugins exported and loaded on RHDH 1.10 (Backstage 1.49.4) — verified in browser (tested via rhdh-local)
- [x] Backward compatibility confirmed: all 5 plugins load on RHDH 1.9 (Backstage 1.45.3) — verified in browser (tested via rhdh-local)
- [x] CI passes on Node 22 and Node 24

Resolves: AAP-78453

## Summary by CodeRabbit

* **New Features**
  * Switched to updated UI stylesheet and improved responsive grid layout for catalog entity pages (better small-screen support).

* **Chores**
  * Expanded supported Node.js runtimes to 22 and 24.
  * Updated Backstage framework and package dependency versions.
  * Bumped application version to 1.49.4.

* **Bug Fixes**
  * Relaxed input/output validation for several scaffolder actions to accept broader payload shapes.

PR #476, opened 2026-07-16
[AAP-83143] fix: restore @backstage/plugin-catalog-backend to ^3.6.1

sshedmake:
## Summary

- Restores `@backstage/plugin-catalog-backend` minimum from `^3.5.0` back to `^3.6.1` in `packages/backend/package.json` and `plugins/catalog-backend-module-rhaap/package.json`
- The Backstage 1.49.4 bulk bump (PR #392) silently downgraded this dependency, reintroducing the `getProcessableEntities()` deadlock (SELECT FOR UPDATE without transaction) fixed in AAP-73301 / PR #330
- yarn.lock now resolves to 3.8.1
- Adds a CI guard (`scripts/check-dependency-pins.sh`) that asserts minimum resolved versions in `yarn.lock` for critical dependencies, preventing future silent downgrades from `backstage-cli versions:bump`

## Root Cause

PR #392 ("Bump Backstage to 1.49.4 & add node.js 24 support") used a bulk version-bump that reset every `@backstage/*` dependency to the versions declared by the 1.49.4 release train, without preserving the manual override from PR #330.

## CI Guard

To add more guarded dependencies in the future, append a line to the `PINNED_DEPS` array in `scripts/check-dependency-pins.sh`:

```bash
PINNED_DEPS=(
  "@backstage/plugin-catalog-backend 3.6.0 AAP-73301:getProcessableEntities-deadlock"
  # Add new entries here:
  # "@backstage/some-package 1.2.3 AAP-XXXXX:short-reason"
)
```

## References

- Regression: [AAP-83143](https://redhat.atlassian.net/browse/AAP-83143)
- Original fix: [AAP-73301](https://redhat.atlassian.net/browse/AAP-73301) / PR #330
- Regressing PR: #392

## Test plan

- [x] Verify `yarn.lock` resolves `@backstage/plugin-catalog-backend` to >= 3.6.0 (resolves to 3.8.1)
- [x] TypeScript type check passes (`yarn tsc`)
- [x] No direct source imports from `@backstage/plugin-catalog-backend` — module uses `plugin-catalog-node/alpha` extension points
- [x] No peer dependency conflicts
- [x] CI guard passes on current lockfile, correctly fails on simulated 3.5.0 regression
- [ ] CI passes
- [ ] No `ConflictError` in catalog processing logs after deployment

Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Maintenance**
  * Updated the backend catalog integration dependency to version 3.6.1.

* **Quality Improvements**
  * Added automated checks to verify required dependency versions before linting and tests run.
  * CI now reports failures when pinned dependencies are missing or below their minimum supported versions.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -31,7 +31,7 @@
     "@backstage/plugin-auth-backend-module-gitlab-provider": "^0.4.1",
     "@backstage/plugin-auth-backend-module-guest-provider": "^0.2.17",
     "@backstage/plugin-auth-node": "^0.6.14",
-    "@backstage/plugin-catalog-backend": "^3.5.0",
+    "@backstage/plugin-catalog-backend": "^3.6.1",
     "@backstage/plugin-catalog-backend-module-logs": "^0.1.20",
     "@backstage/plugin-catalog-backend-module-scaffolder-entity-model": "^0.2.18",
     "@backstage/plugin-kubernetes-backend": "^0.21.2",
--- a/scripts/check-dependency-pins.sh
+++ b/scripts/check-dependency-pins.sh
@@ -3,6 +3,9 @@
 # Add entries to PINNED_DEPS when a manual override fixes a critical bug that the
 # Backstage release train doesn't carry at a high-enough version.
 #
+# Checks EVERY resolved version block for the package in yarn.lock, not just the
+# first match — multiple coexisting ranges per package are the norm in this repo.
+#
 # Format: "package minimum_major.minor.patch reason"
 PINNED_DEPS=(
   "@backstage/plugin-catalog-backend 3.6.0 AAP-73301:getProcessableEntities-deadlock"
@@ -12,24 +15,25 @@ rc=0
 for entry in "${PINNED_DEPS[@]}"; do
   read -r pkg min_ver reason <<< "$entry"
 
-  resolved=$(grep -A1 "\"${pkg}@npm:" yarn.lock \
+  mapfile -t versions < <(grep -A1 "\"${pkg}@npm:" yarn.lock \
     | grep 'version:' \
-    | head -1 \
     | sed 's/.*version: //')
 
-  if [ -z "$resolved" ]; then
+  if [ ${#versions[@]} -eq 0 ]; then
     echo "FAIL: ${pkg} not found in yarn.lock"
     rc=1
     continue
   fi
 
-  lowest=$(printf '%s\n%s\n' "$min_ver" "$resolved" | sort -V | head -1)
-  if [ "$lowest" != "$min_ver" ]; then
-    echo "FAIL: ${pkg} resolved to ${resolved}, minimum required is ${min_ver} (${reason})"
-    rc=1
-  else
-    echo "OK: ${pkg}@${resolved} >= ${min_ver} (${reason})"
-  fi
+  for resolved in "${versions[@]}"; do
+    lowest=$(printf '%s\n%s\n' "$min_ver" "$resolved" | sort -V | head -1)
+    if [ "$lowest" != "$min_ver" ]; then
+      echo "FAIL: ${pkg} resolved to ${resolved}, minimum required is ${min_ver} (${reason})"
+      rc=1
+    else
+      echo "OK: ${pkg}@${resolved} >= ${min_ver} (${reason})"
+    fi
+  done
 done
 
 exit $rc
--- a/scripts/check-dependency-pins.sh
+++ b/scripts/check-dependency-pins.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+# Guard against silent dependency downgrades from bulk backstage-cli versions:bump.
+# Add entries to PINNED_DEPS when a manual override fixes a critical bug that the
+# Backstage release train doesn't carry at a high-enough version.
+#
+# Format: "package minimum_major.minor.patch reason"
+PINNED_DEPS=(
+  "@backstage/plugin-catalog-backend 3.6.0 AAP-73301:getProcessableEntities-deadlock"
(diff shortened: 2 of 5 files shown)


Jeevansuresh1 (approved):

scripts/check-dependency-pins.sh
```diff
@@ -0,0 +1,35 @@
...
+)
+
+rc=0
+for entry in "${PINNED_DEPS[@]}"; do
+  read -r pkg min_ver reason <<< "$entry"
+
+  resolved=$(grep -A1 "\"${pkg}@npm:" yarn.lock \
+    | grep 'version:' \
+    | head -1 \
+    | sed 's/.*version: //')
```
NilashishC: Nice guard, but the lockfile parsing here can pick an arbitrary matching block rather than the one the guarded package actually resolves to.

`grep "\"${pkg}@npm:"` matches *every* version-range block for that package name, not just the range declared by the packages we care about. I checked the current lockfile and there are two separate blocks for this exact package:

```
"@backstage/plugin-catalog-backend@npm:^3.6.1":
  version: 3.8.1
"@backstage/plugin-catalog-backend@npm:^3.7.0":
  version: 3.7.1
```

`head -1` happens to grab `3.8.1` here only because `^3.6.1` sorts before `^3.7.0` textually in the lockfile -- that's coincidental, not guaranteed. If a future divergent range for this package (or any future `PINNED_DEPS` entry) sorts earlier and resolves below the minimum, this check would report `OK` while the guarded package is still on a bad version -- the same silent-regression failure mode this script exists to prevent.

This isn't hypothetical for this repo: `@backstage/backend-openapi-utils`, `@backstage/backend-plugin-api`, and `@backstage/plugin-catalog-node` all currently have 3-4 coexisting divergent-version blocks in `yarn.lock`, so multi-resolution per package name is the norm here, not an edge case.

Suggestion: scope the check to the exact range declared in the guarded package's own `package.json` (read the range for `packages/backend` / `plugins/catalog-backend-module-rhaap`, then grep the lockfile for the literal quoted key `"${pkg}@npm:${range}":` instead of a name-prefix match), or use `yarn info ${pkg} --json` scoped per workspace.

packages/backend/package.json
```diff
@@ -31,7 +31,7 @@
     "@backstage/plugin-auth-backend-module-gitlab-provider": "^0.4.1",
     "@backstage/plugin-auth-backend-module-guest-provider": "^0.2.17",
     "@backstage/plugin-auth-node": "^0.6.14",
-    "@backstage/plugin-catalog-backend": "^3.5.0",
+    "@backstage/plugin-catalog-backend": "^3.6.1",
```
NilashishC, line 34: This fix targets `main` only. Worth noting: `release-2.2` has the same regression via a separate backport commit (`c2afdf5f`, PR #438, tracked as AAP-81959), which isn't an ancestor of this branch. `release-2.2`'s `v2.2.4` tag (Jira `portal-2.2.4`) is targeted for release today, so it currently carries this deadlock forward. A companion backport PR against `release-2.2` would be worth opening given the timing -- not a blocker for merging this PR into `main`, just flagging so it doesn't get missed.

NilashishC (commented):
Core fix looks correct -- verified in a worktree that yarn.lock now resolves @backstage/plugin-catalog-backend to 3.8.1 for both packages/backend and plugins/catalog-backend-module-rhaap, above the 3.6.0 minimum from AAP-73301. Two things worth a look, left inline: the CI guard's lockfile parsing has a latent gap that could let a similar regression slip through silently in the future, and release-2.2 (already tagged as v2.2.4, targeted for release today) has the same regression via a separate backport and isn't covered by this PR.

packages/backend/package.json
sshedmake: Agreed, will open a companion backport PR against release-2.2 once this merges into main.

scripts/check-dependency-pins.sh
sshedmake: Good catch. Fixed in `9cb5953` — the script now iterates over every resolved version block for the package instead of head -1. Verified it catches a mixed lockfile where one range resolves to 3.8.1 (OK) and 
another to 3.5.0 (FAIL).

sshedmake: **Update**

Addressed fully in `2a86ec00` — the script now reads the declared range from each package.json under packages/ and plugins/, then checks the exact resolution for that range in yarn.lock (i.e., 
`"${pkg}@npm:${range}":` match). This prevents false-OKs from unrelated transitive ranges resolving above the minimum while the guarded packages resolve below it.

Verified with a simulated regression where `packages/backend` declares ^3.5.0 (resolves to 3.5.0) and `catalog-backend-module-rhaap` declares ^3.6.1 (resolves to 3.8.1) — the script now correctly fails on the 
first and passes on the second, with output pinpointing the offending file.

NilashishC (approved):
Re-reviewed after the latest commits. Independently verified the check-dependency-pins.sh fix, not just read the diff: ran it against the real lockfile (passes cleanly), then simulated the exact regression scenario by reverting packages/backend's declared range back to ^3.5.0 -- the script correctly fails with a clear file-attributed message, while catalog-backend-module-rhaap still correctly passes. The exact-range-match approach resolves the coincidental-ordering gap I flagged earlier. Resolved that thread.

CodeRabbit's mapfile/CRLF comment is moot against the final version -- the script no longer uses mapfile at all.

The release-2.2 backport note is still open but non-blocking for this PR into main; you've already committed to a companion PR, which is the right call given today's v2.2.4 timing.

Approving.

audgirka (approved):

abhikdps (approved):

PR #508, opened 2026-07-30
feat(e2e): add RBAC role creation and catalog sync to global setup

Jeevansuresh1:
## Summary

- Adds `rbac-setup.ts` utility that launches a browser, logs in as admin, triggers catalog sync via the "Sync now" UI dialog, and walks the 4-step RBAC wizard to create a role with all plugin permissions for the non-admin test user
- Updates `global-setup.ts` to call `setupNonAdminRBAC()` after `createNonAdminTestUser()`, so non-admin e2e tests have proper RBAC permissions before execution
- Role creation is idempotent (skips if the role already exists) and configurable via `RBAC_ROLE_NAME` and `AAP_NONADMIN_USER_ID` env vars

## Test plan

- [ ] Verify `globalSetup` runs successfully with `AAP_NONADMIN_USER_ID` and `AAP_TOKEN` set
- [ ] Confirm catalog sync triggers before RBAC role creation
- [ ] Confirm RBAC role is created with all plugins and permissions for the non-admin user
- [ ] Verify idempotency: re-running global setup skips role creation if role already exists
- [ ] Verify non-admin tests (`non-admin-pages`, `non-admin-permissions`, `non-admin-templates`) pass with the new setup

### DIFF
--- a/e2e-tests/playwright/global-setup.ts
+++ b/e2e-tests/playwright/global-setup.ts
@@ -1,15 +1,11 @@
 import { createNonAdminTestUser } from './utils/aap-user-setup';
+import { setupNonAdminRBAC } from './utils/rbac-setup';
 
+/** Creates the non-admin AAP user, syncs the catalog, and creates an RBAC role via the UI wizard. */
 async function globalSetup() {
   if (process.env.AAP_NONADMIN_USER_ID && process.env.AAP_TOKEN) {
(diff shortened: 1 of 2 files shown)


Jeevansuresh1:
https://github.com/ansible/ansible-backstage-plugins/pull/509

PR #510, opened 2026-07-31
fix: cleanup stale form data when unchecking conditional checkboxes in EE scaffolder

Abhishekism9450:
## Description

When users uncheck conditional checkboxes in the EE scaffolder template (e.g., "Publish to Git repository", "Build Execution Environment"), the dependent form fields remain in the form data. This causes execution failures when the backend receives data for features that were explicitly unchecked.

## Solution

Implemented schema-based form data cleanup that automatically removes dependent fields when conditional checkboxes are unchecked.

Key Features:

- Generic solution: Works for any conditional checkbox by detecting JSON Schema dependencies
- Distinguishes field types: Conditional checkboxes (have dependencies) trigger cleanup, simple booleans do not
- Step-isolated: Only cleans current step's data, preserves values from other wizard steps
- Key-scoped: Only cleans the specific nested object that had a toggle, not sibling objects
- Defensive: Runs in both onChange (field changes) and onSubmit (navigation) to catch edge cases

How It Works:

1. fieldHasDependenciesInSchema() - Recursively searches schema tree (including nested oneOf branches) to determine if a field has dependencies
2. getActiveSchemaProperties() - Parses JSON Schema to determine which properties should exist based on current form values
3. cleanupFormDataAgainstSchema() - Removes fields not in the active schema, recursively handles nested objects
4. cleanupNestedFields() - Detects boolean toggles (true → false) for conditional checkboxes, tracks which specific key toggled, and triggers cleanup only for that key

## Related Issues

https://redhat.atlassian.net/browse/AAP-84907

## Checklist

- [ ] Code follows project style
- [ ] Tests pass locally
- [ ] Documentation updated

## Bug Fixes
- Inactive conditional form fields now have their nested values removed correctly.
- Unrelated form data is preserved when conditional options change.
- Nested dependencies, rapid toggling, invalid inputs, and simultaneous updates are handled more reliably.
- Form submissions now consistently exclude values from inactive conditional sections.

## Tests
- Added comprehensive coverage for conditional visibility, dependency handling, recursive cleanup, batch updates, and data preservation.

### DIFF
--- a/plugins/self-service/src/components/CreateTask/StepForm.tsx
+++ b/plugins/self-service/src/components/CreateTask/StepForm.tsx
@@ -446,6 +446,144 @@ function isPlainObject(value: unknown): value is Record<string, any> {
   return value !== null && typeof value === 'object' && !Array.isArray(value);
 }
 
+/**
+ * Determines which properties should be active in the schema based on current form values.
+ * Handles JSON Schema dependencies with oneOf branches.
+ */
+export function getActiveSchemaProperties(
+  schema: Record<string, any>,
+  currentFormData: Record<string, any>,
+): Set<string> {
+  const activeProps = new Set<string>();
+
+  if (!schema?.properties) {
+    return activeProps;
+  }
+
+  // Add base properties
+  Object.keys(schema.properties).forEach(prop => activeProps.add(prop));
+
+  // Process dependencies
+  if (schema.dependencies) {
+    for (const [depKey, depValue] of Object.entries(schema.dependencies)) {
+      if (!depValue || typeof depValue !== 'object') continue;
+
+      const depValueObj = depValue as any;
+      if (Array.isArray(depValueObj.oneOf)) {
+        for (const branch of depValueObj.oneOf) {
+          if (!branch.properties) continue;
+
+          // Check if this branch matches
+          const branchProp = branch.properties[depKey];
+          if (branchProp && typeof branchProp === 'object' && 'const' in branchProp) {
+            if (currentFormData[depKey] === branchProp.const) {
+              // Branch is active, add its properties
+              Object.keys(branch.properties).forEach(prop => {
+                if (prop !== depKey) {
+                  activeProps.add(prop);
+                }
+              });
+            }
+          }
+        }
+      }
+    }
+  }
+
+  return activeProps;
+}
+
+/**
+ * Recursively checks if a field name has dependencies defined anywhere in the schema tree.
+ * Used to distinguish conditional checkboxes (have dependencies) from simple boolean fields (no dependencies).
+ */
+export function fieldHasDependenciesInSchema(fieldName: string, schema: Record<string, any>): boolean {
+  if (!schema || typeof schema !== 'object') {
+    return false;
+  }
+
+  // Check if dependencies exist at this level
+  if (schema.dependencies?.[fieldName]) {
+    return true;
+  }
+
+  // Recursively check in oneOf branches
+  if (Array.isArray(schema.oneOf)) {
+    for (const branch of schema.oneOf) {
+      if (fieldHasDependenciesInSchema(fieldName, branch)) {
+        return true;
+      }
+    }
+  }
+
+  // Recursively check in dependencies oneOf branches
+  if (schema.dependencies) {
+    for (const depValue of Object.values(schema.dependencies)) {
+      if (typeof depValue === 'object' && depValue !== null) {
+        const depValueObj = depValue as any;
+        if (Array.isArray(depValueObj.oneOf)) {
+          for (const branch of depValueObj.oneOf) {
+            if (fieldHasDependenciesInSchema(fieldName, branch)) {
+              return true;
+            }
+          }
+        }
+      }
+    }
+  }
+
+  // Recursively check in properties
+  if (schema.properties) {
+    for (const propValue of Object.values(schema.properties)) {
+      if (typeof propValue === 'object' && propValue !== null) {
+        if (fieldHasDependenciesInSchema(fieldName, propValue as Record<string, any>)) {
+          return true;
+        }
+      }
+    }
+  }
+
+  return false;
+}
+
+/**
+ * Cleans form data against schema, removing fields that are not active based on dependencies.
+ * Recursively cleans nested objects.
+ */
+export function cleanupFormDataAgainstSchema(
+  currentFormData: Record<string, any>,
+  schema: Record<string, any>,
+): Record<string, any> {
+  if (!schema || !currentFormData || typeof currentFormData !== 'object') {
+    return currentFormData;
+  }
+
+  const cleaned: Record<string, any> = {};
+  const activeProps = getActiveSchemaProperties(schema, currentFormData);
+
+  for (const key of Object.keys(currentFormData)) {
+    const value = currentFormData[key];
+
+    if (!activeProps.has(key)) {
+      continue;
+    }
+
+    // Recursively clean nested objects
+    if (value && typeof value === 'object' && !Array.isArray(value)) {
+      const nestedSchema = schema.properties?.[key];
+      if (nestedSchema?.properties || nestedSchema?.dependencies) {
+        cleaned[key] = cleanupFormDataAgainstSchema(value, nestedSchema);
+      } else {
+        cleaned[key] = value;
+      }
+    } else {
+      cleaned[key] = value;
+    }
+  }
+
+  return cleaned;
+}
+
 export function deepMergePlainObjects(
   base: Record<string, any>,
   patch: Record<string, any>,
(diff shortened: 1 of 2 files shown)


plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -668,6 +806,78 @@ export const StepForm = ({
     setActiveStep(prevActiveStep => prevActiveStep - 1);
   };
 
+  // Get all active properties based on current form data and schema dependencies
```
NilashishC: Nit: This comment appears orphaned — it doesn't precede any corresponding code (the next line is blank, then another comment + `cleanupNestedFields`). Looks like a leftover from refactoring. Consider removing it or merging it into the comment on line 811.

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -668,6 +806,78 @@ export const StepForm = ({
...
+
+          if (checkNestedToggle(currentValue, prevValue || {})) {
+            hasToggleInCurrentStep = true;
+            break;
+          }
+        }
+      }
+
+      // Only run cleanup if toggle was detected in current step
+      if (hasToggleInCurrentStep) {
```
NilashishC: When any nested object in the step has a qualifying toggle, **all** nested-object properties in the step get cleaned (the loop iterates `stepPropertyKeys`), not just the property that contains the toggle.

If a step had two sibling nested-object properties (e.g. `publishAndBuild` and `gitConfig`), a toggle in one would unnecessarily re-clean the other. Consider scoping cleanup to only the `topLevelKey` where the toggle was detected, or at least breaking out of the outer loop once you have the offending key and cleaning only that subtree.

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -680,15 +890,12 @@ export const StepForm = ({
...
+      setFormData(prev => {
+        const merged = mergeStepFormDataHybrid(prev, step, data.formData as Record<string, any>);
+        return cleanupNestedFields(merged, prev, stepIndex);
+      });
     },
-    [filteredSteps],
+    [filteredSteps, cleanupNestedFields],
   );
 
   const handleFormSubmit = useCallback(
```
NilashishC, line 948: `handleFormChange` applies `cleanupNestedFields` after `mergeStepFormDataHybrid`, but `handleFormSubmit` here does not. In normal flow RJSF fires `onChange` before `onSubmit` so stale data would already be cleaned, but for consistency (and to guard against edge cases where `onSubmit` fires without a preceding `onChange`), consider applying the same cleanup here:

```suggestion
  const handleFormSubmit = useCallback(
    (stepIndex: number, data: IChangeEvent<any>) => {
      if (data.formData) {
        const step = filteredSteps[stepIndex];
        if (step) {
          setFormData(prev => {
            const merged = mergeStepFormDataHybrid(
              prev,
              step,
              data.formData as Record<string, any>,
            );
            return cleanupNestedFields(merged, prev, stepIndex);
          });
        }
      }
      handleNext();
    },
    [filteredSteps, handleNext, cleanupNestedFields],
  );
```

plugins/self-service/src/components/CreateTask/StepForm.tsx
NilashishC, line 1: Minor: The file is missing a trailing newline (diff shows `\ No newline at end of file`). This will likely trigger Prettier/EditorConfig. Easy fix.

NilashishC (commented):
Nice fix — the schema-driven cleanup approach is well-designed. The distinction between conditional checkboxes (have dependencies) and simple booleans via `fieldHasDependenciesInSchema` is elegant. Tests are thorough.

A few minor things to address in the inline comments (orphaned comment, missing trailing newline, cleanup consistency between onChange/onSubmit, and scope of the cleanup loop). Nothing blocking — overall this looks good.

plugins/self-service/src/components/CreateTask/StepForm.tsx
Abhishekism9450: Fixed

plugins/self-service/src/components/CreateTask/StepForm.tsx
Abhishekism9450: Good catch.  Refactored the code to track toggle key and clean up those only.

plugins/self-service/src/components/CreateTask/StepForm.tsx
Abhishekism9450: Done. Modified the code as suggested.

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -668,6 +818,97 @@ export const StepForm = ({
...
+                }
+              }
+            }
+            return false;
+          };
+
+          if (checkNestedToggle(currentValue, prevValue || {})) {
+            toggledKey = topLevelKey;
+            break;
+          }
```
abhikdps, line 893: Nit: This `break` means only one top-level key gets cleaned per change event. If two conditional checkboxes under different top-level keys toggle `true→false` simultaneously (e.g., a programmatic update or a future batch state restore), only the first detected toggle's dependents are cleaned — the second key's stale fields survive.

Low practical risk since normal user interaction fires one `onChange` per checkbox click, but if you want to be defensive you could collect all toggled keys instead of breaking on the first:

```ts
const toggledKeys: string[] = [];
for (const topLevelKey of stepPropertyKeys) {
  // ... existing detection logic ...
  if (checkNestedToggle(currentValue, prevValue || {})) {
    toggledKeys.push(topLevelKey);
    // no break — keep scanning
  }
}

if (toggledKeys.length > 0) {
  const cleaned = { ...merged };
  for (const key of toggledKeys) {
    const propSchema = step.schema.properties[key];
    if (cleaned[key] && typeof cleaned[key] === 'object' && !Array.isArray(cleaned[key]) && propSchema) {
      cleaned[key] = cleanupFormDataAgainstSchema(cleaned[key], propSchema);
    }
  }
  return cleaned;
}
```

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -446,6 +446,144 @@ function isPlainObject(value: unknown): value is Record<string, any> {
...
+
+    if (!activeProps.has(key)) {
+      continue;
+    }
+
+    // Recursively clean nested objects
+    if (value && typeof value === 'object' && !Array.isArray(value)) {
+      const nestedSchema = schema.properties?.[key];
+      if (nestedSchema?.properties || nestedSchema?.dependencies) {
+        cleaned[key] = cleanupFormDataAgainstSchema(value, nestedSchema);
```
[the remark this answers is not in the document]
Abhishekism9450: Good catch. Addressed in https://github.com/ansible/ansible-backstage-plugins/commit/79972748e4990a2d9ef074944823b3be455013b9

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -446,6 +446,144 @@ function isPlainObject(value: unknown): value is Record<string, any> {
...
+                if (prop !== depKey) {
+                  activeProps.add(prop);
+                }
+              });
+            }
+          }
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Abhishekism9450: Thanks for the suggestion! Our current templates exclusively use const-based discriminators (verified in eeTemplate.ts). Adding enum support would be future-proofing for a case we don't currently have.

For allOf/if/then conditionals, our schema structure uses dependencies.oneOf exclusively, so those branches aren't currently needed.

plugins/self-service/src/components/CreateTask/StepForm.tsx
```diff
@@ -668,6 +806,78 @@ export const StepForm = ({
...
+        for (const key of stepPropertyKeys) {
+          if (cleaned[key] && typeof cleaned[key] === 'object' && !Array.isArray(cleaned[key])) {
+            const propSchema = step.schema.properties[key];
+            if (propSchema) {
+              cleaned[key] = cleanupFormDataAgainstSchema(cleaned[key], propSchema);
+            }
+          }
+        }
+        return cleaned;
+      }
```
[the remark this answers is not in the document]
Abhishekism9450: Our schema structure guarantees conditional checkboxes are always nested inside step-level objects.

Root-level conditional booleans are not used in our schema design, so the scenario you described (conditional checkbox as a direct step property) doesn't occur.

If we ever change the schema design to include root-level conditional booleans, we can add support for that case. For now, the current implementation correctly handles our schema structure.

@NilashishC  should we consider this now or not ?

plugins/self-service/src/components/CreateTask/StepForm.tsx
Abhishekism9450: Great catch! Fixed. 
The early break meant only the first detected toggle got cleaned. This would cause stale data to survive when multiple conditional checkboxes toggle simultaneously.

abhikdps:
@coderabbitai review

abhikdps (approved):

PR #536, opened 2026-08-09
feat(auth): add PKCE support for AAP 2.7 OAuth compliance

Jeevansuresh1:
## Summary

AAP 2.7 enforces PKCE (Proof Key for Code Exchange, RFC 7636) on all OAuth authorization flows. The Portal's auth backend module currently generates OAuth authorize URLs without `code_challenge` parameters, causing AAP to reject every login request with `"This application requires PKCE"`. This breaks all browser-based authentication.

**Changes:**

- **authenticator.ts**: Generate a cryptographically random `code_verifier` and its SHA-256 `code_challenge` during the OAuth `start()` phase. Store the verifier keyed by OAuth state in an in-memory Map with a 10-minute TTL. Append `code_challenge` and `code_challenge_method=S256` to the authorization URL. Retrieve and pass the verifier during `authenticate()` for the token exchange. Fail explicitly with a descriptive error if the PKCE entry is missing (expired session or server restart).
- **AAPClient.ts**: Accept an optional `codeVerifier` parameter in `rhAAPAuthenticate()` and include `code_verifier` in the token exchange POST body when present.
- **module.test.ts**: Updated the `should start` test to expect `code_challenge` and `code_challenge_method` in the authorization URL parameters.

## Verified in CI

Tested in Jenkins pipeline (build #1256) using the `on-pr-802` image after updating the backstage-plugins submodule in ansible-rhdh-plugins. Pipeline passed with 96.05% pass rate and 0 Playwright failures.

## Test plan

- [x] `yarn tsc` passes
- [x] `yarn lint` passes
- [x] `yarn test` passes (187 suites, 3945 tests)
- [x] Jenkins pipeline build runs successfully 

## Summary by CodeRabbit

- **Security**
  - Added enhanced protection for AAP OAuth login with PKCE validation.
  - OAuth callbacks now reject missing, unknown, or expired authorization states.

- **Bug Fixes**
  - Improved login error handling so failed AAP credential submissions are reported clearly.

- **Documentation**
  - Added deployment guidance for OAuth flows, including sticky-session requirements for multi-replica setups.

### DIFF
--- a/plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
+++ b/plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
@@ -10,6 +10,7 @@ import { IAAPService } from '@ansible/backstage-rhaap-common';
 import { AuthenticationError } from '@backstage/errors';
 
 const PKCE_TTL_MS = 10 * 60 * 1000;
+const PKCE_MAX_ENTRIES = 10_000;
 const pkceStore = new Map<string, { verifier: string; createdAt: number }>();
 
 function generatePKCE(): { verifier: string; challenge: string } {
@@ -26,6 +27,10 @@ function cleanExpiredPKCE(): void {
     if (now - entry.createdAt > PKCE_TTL_MS) {
       pkceStore.delete(key);
     }
+  }
+  while (pkceStore.size > PKCE_MAX_ENTRIES) {
+    const oldest = pkceStore.keys().next().value;
+    if (oldest) pkceStore.delete(oldest);
   }
 }
 
@@ -105,7 +110,12 @@ export const aapAuthAuthenticator = (aapService: IAAPService) =>
       input,
       { host, clientId, clientSecret, callbackURL, checkSSL },
     ) {
-      const state = input.req.query.state as string;
+      const state = input.req.query.state as string | undefined;
+      if (!state) {
+        throw new Error(
+          'OAuth state parameter missing from callback request.',
+        );
+      }
       const pkceEntry = pkceStore.get(state);
       if (!pkceEntry) {
         throw new Error(
--- a/plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
+++ b/plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
@@ -1,3 +1,4 @@
+import { randomBytes, createHash } from 'node:crypto';
 import { Strategy as OAuth2Strategy } from 'passport-oauth2';
 import {
   createOAuthAuthenticator,
@@ -7,6 +8,26 @@ import {
 } from '@backstage/plugin-auth-node';
 import { IAAPService } from '@ansible/backstage-rhaap-common';
 import { AuthenticationError } from '@backstage/errors';
+
+const PKCE_TTL_MS = 10 * 60 * 1000;
+const pkceStore = new Map<string, { verifier: string; createdAt: number }>();
+
+function generatePKCE(): { verifier: string; challenge: string } {
+  const verifier = randomBytes(32).toString('base64url');
+  const challenge = createHash('sha256')
+    .update(verifier)
+    .digest('base64url');
+  return { verifier, challenge };
+}
+
+function cleanExpiredPKCE(): void {
+  const now = Date.now();
+  for (const [key, entry] of pkceStore) {
+    if (now - entry.createdAt > PKCE_TTL_MS) {
+      pkceStore.delete(key);
+    }
+  }
+}
 
 /** @public */
 export interface AAPAuthenticatorContext {
@@ -64,12 +85,19 @@ export const aapAuthAuthenticator = (aapService: IAAPService) =>
       return { helper, host, clientId, clientSecret, callbackURL, checkSSL };
     },
     async start(input, { helper }) {
+      const { verifier, challenge } = generatePKCE();
+      cleanExpiredPKCE();
+      pkceStore.set(input.state, {
+        verifier,
+        createdAt: Date.now(),
+      });
+
       const start = await helper.start(input, {
         accessType: 'offline',
         prompt: 'auto',
         approval_prompt: 'auto',
       });
-      start.url += '&approval_prompt=auto';
+      start.url += `&approval_prompt=auto&code_challenge=${challenge}&code_challenge_method=S256`;
       return start;
     },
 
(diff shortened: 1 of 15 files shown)


sshedmake (approved):
**LGTM — PKCE implementation is clean and RFC 7636-compliant.**

Verified locally:
- `yarn tsc` passes
- All auth plugin tests pass (3 suites, 24 tests)
- Code review scored 10/10 across Functionality, Security, and Quality lenses

Implementation highlights:
- Cryptographically random verifier (`randomBytes(32)`), S256 challenge
- Single-use verifier consumption with 10-min TTL and 10K entry cap
- Backward-compatible — `codeVerifier` is optional, older AAPs simply ignore the extra params
- Error messages guide users to retry on session expiry or server restart

Minor suggestions (non-blocking):
- Consider adding a happy-path unit test for `authenticate()` with valid PKCE state (currently covered by `module.test.ts` integration test)
- Consider a boundary test at `TTL - 1ms` to confirm the near-expiry happy path

Deployment note: the in-memory PKCE store is process-local — multi-replica setups need sticky sessions or the callback may hit a different pod. Worth documenting if not already covered.

plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
```diff
@@ -64,26 +85,43 @@ export const aapAuthAuthenticator = (aapService: IAAPService) =>
...
       { host, clientId, clientSecret, callbackURL, checkSSL },
     ) {
+      const state = input.req.query.state as string;
+      const pkceEntry = pkceStore.get(state);
+      if (!pkceEntry) {
+        throw new Error(
+          'PKCE verifier not found for OAuth state. The login session may have expired or the server may have restarted. Please try logging in again.',
+        );
+      }
+      pkceStore.delete(state);
```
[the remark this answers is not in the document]
Jeevansuresh1: Fixed in commit 7a732a5. The authenticate callback now checks Date.now() - pkceEntry.createdAt >= PKCE_TTL_MS and rejects expired verifiers before the token exchange. A callback-after-TTL test was also added.

plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
```diff
@@ -64,26 +88,51 @@ export const aapAuthAuthenticator = (aapService: IAAPService) =>
...
+      const state = input.req.query.state as string | undefined;
+      if (!state) {
+        throw new Error('OAuth state parameter missing from callback request.');
+      }
+      const pkceEntry = pkceStore.get(state);
+      pkceStore.delete(state);
+      if (!pkceEntry) {
+        throw new Error(
+          'PKCE verifier not found for OAuth state. The login session may have expired or the server may have restarted. Please try logging in again.',
+        );
```
NilashishC, line 120: Nit / docs hook: this error already covers expiry and process restart, but in HA Portal deployments the same symptom also happens when the OAuth callback lands on a different replica than `start` (process-local `pkceStore`).

Please either mention that possibility here, or — preferably — document the sticky-session / single-replica requirement in `docs/features/external-authentication.md` and the auth provider README before merge. The open CodeRabbit thread on the in-memory store is still valid; documenting the constraint is the minimum bar if shared/cookie-backed storage is deferred.

NilashishC (changes requested):
PKCE implementation looks solid and well tested — thanks for unblocking AAP 2.7 login.

Requesting changes mainly for docs/ops readiness: the in-memory verifier store (already flagged and still open) is a real risk for Portal HA deployments (`spec.backstage.replicas` / HPA), and this PR does not yet document the sticky-session / single-replica constraint. Please update `docs/features/external-authentication.md` and the auth provider README accordingly; a follow-up for cookie- or shared-store PKCE is fine if you want to keep this PR focused.

NilashishC:
### Follow-up: prefer a cookie-bound PKCE verifier

A better replica-safe approach is **#1: cookie-bound verifier** (same idea as the existing `rhaap-nonce` cookie): put the short-lived secret in the browser, not in pod memory.

#### Flow

```mermaid
sequenceDiagram
  participant Browser
  participant PodA as Backstage pod A
  participant PodB as Backstage pod B
  participant AAP

  Browser->>PodA: GET /api/auth/rhaap/start
  Note over PodA: generate verifier + S256 challenge
  PodA-->>Browser: 302 to AAP + Set-Cookie: rhaap-pkce=verifier
  Browser->>AAP: /o/authorize/?code_challenge=...&state=...
  AAP-->>Browser: redirect with ?code&state
  Browser->>PodB: GET /handler/frame?code&state (sends cookie)
  Note over PodB: read verifier from cookie, clear cookie
  PodB->>AAP: POST /o/token/ with code + code_verifier
```

#### What would change

**`start`**
1. Generate verifier + challenge (as today).
2. Instead of `pkceStore.set(state, …)`, `Set-Cookie: rhaap-pkce=<verifier>` with `HttpOnly; Secure; SameSite=Lax; Path=/api/auth/rhaap; Max-Age=600`.
3. Still append `code_challenge` / `S256` to the authorize URL.

**`authenticate`**
1. Read verifier from the cookie.
2. Clear the cookie (`Max-Age=0`) so it is single-use.
3. Optionally bind it to `state` so a leftover cookie from another attempt cannot be reused.
4. Pass verifier into `rhAAPAuthenticate` as today.
5. Drop the in-memory `Map`.

#### Why this works across replicas

The browser stores the verifier. Whichever pod handles the callback gets the cookie on that request — no shared server state.

#### Safety knobs

| Property | How |
|---|---|
| Not XSS-readable | `HttpOnly` |
| HTTPS only in prod | `Secure` |
| Sent on OAuth return | `SameSite=Lax` (top-level redirect from AAP) |
| Short-lived | `Max-Age` ≈ 10 minutes |
| Single-use | Clear cookie on callback |
| Tied to this login | Include `state` in the cookie value / signature |

#### Practical note

`createOAuthAuthenticator`’s `start` return is mainly `{ url, … }`. We may need to set the cookie on the response object Backstage exposes in that hook (same place the framework sets `rhaap-nonce`). If the authenticator API can’t set cookies directly, the fallback is to embed the verifier in the **signed OAuth state** (alongside `nonce`) — also replica-safe, still no Redis.

**Suggestion for this PR:** document sticky-session / single-replica as the short-term constraint; follow up with cookie (or signed-state) PKCE rather than introducing Redis.

plugins/auth-backend-module-rhaap-provider/src/authenticator.ts
Jeevansuresh1: I've updated the PKCE error message to mention multi-replica routing and added deployment notes about the sticky session requirement in the auth provider README and external-authentication docs.

NilashishC (approved):
Approving this based on the understanding that a follow-up PR will be opened that makes this multi-replica safe.

Abhishekism9450 (approved):

abhikdps (approved):

audgirka (approved):

PR #537, opened 2026-08-10
feat(apme): replace settings allowlist with RBAC settings view/edit permissions

NilashishC:
## Description

Replaces the standalone `ensureSettingsAdmin` config allowlist gate (`ansible.apme.settingsAdminEntityRefs`, fail-open when unset) with shared resource permissions in Backstage RBAC, and adds explicit view-level gating for settings UI/routes.

This moves APME settings authorization to the standard permission-policy framework and scopes access per capability using `FOR_CAPABILITY`, so policy authors can grant narrowly (for now, `apme`) instead of relying on app-config-only allow/deny behavior.

- Move `backstage-rhaap-common/src/permissions.ts` into `permissions/index.ts` and add `permissions/rules.ts` while keeping the public `./permissions` export stable
- Add shared settings resource model:
  - `ansible.settings.edit`
  - `ansible.settings.view`
  - `ansibleSettingsResourceRef`
  - `FOR_CAPABILITY` rule
- Replace mutating-route admin checks with `PermissionsService` middleware (`requireSettingsManage`) for settings/galaxy/AI/rules writes
- Add `requireSettingsView` middleware and gate settings-exclusive reads:
  - `GET /apme/settings/galaxy-servers`
  - `GET /apme/ai/config`
  - `GET /apme/ai/engines`
- Keep shared non-settings reads open (used by non-admin scan/workflow UX):
  - `/apme/settings`, `/apme/ai/status`, `/apme/ai/models`, `/apme/ai/providers`
- Frontend permission integration:
  - Wrap settings content with `RequirePermission` + `resourceRef="apme"`
  - Add tab-level extension gating so unauthorized users do not see the **Quality settings** tab at all (and deep links redirect away instead of clickable-then-404)
  - Gate the **“Quality settings →”** link in Fleet Quality so it is hidden without settings view permission
- Narrow capability list/rule schema to currently wired capability (`apme`) and drop unused `aap`/`general` values
- Add/expand unit tests for:
  - manage/view middleware ALLOW/DENY behavior
  - router gating and shared-read regressions
  - hidden-vs-visible tab/link behavior in frontend
- Keep `catalog-backend-module-apme/tsconfig.json` deprecation handling (`ignoreDeprecations: "5.0"`)
- Ignore local `.data/` runtime dir and `*.tsbuildinfo` build artifacts
- Update `AGENTS.md` RBAC docs for settings view/edit and tab visibility behavior

## Related Issues

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Added test cases.

image

## Checklist

- [x] Tests pass locally
- [x] Documentation updated


ganeshrn (commented):
## Recommendation: make settings capabilities plugin-registered (startup-dynamic)

The `FOR_CAPABILITY` rule currently hardcodes the dropdown via:

```ts
capability: z.enum(['apme', 'aap', 'general'])
```

That works for a first cut, but the capability list should not live as speculative literals in `backstage-rhaap-common`. RHDH RBAC builds Rule Parameters from `paramsSchema` in permission metadata (`GET /api/permission/plugins/condition-rules`), so the enum is inherently fixed at **backend process start** — not fetched live from APME or the RBAC UI.

### Goal

Capabilities that own settings (APME today; AAP / general later) should **register themselves** when their backend module loads. Changing which plugins/modules are enabled in config, then **restarting the backend**, should refresh the `FOR_CAPABILITY` options without a separate hardcoded list.

> Note: this will not hot-reload without a restart. That matches how Backstage registers plugins/services and how permission metadata is published. Restart-after-config-change is the right contract.

### Proposed shape

```text
catalog-module-apme          ──addCapability('apme')──┐
future aap settings module   ──addCapability('aap')───┼──► ansibleSettingsCapabilities
general settings module      ──addCapability('general')┘         extension point
                                                                      │
                                                                      ▼
                                                         settings-permissions module
                                                         (aggregates registrations,
                                                          then addResourceType once)
                                                                      │
                                                                      ▼
                                                         RBAC UI: FOR_CAPABILITY
                                                         enum = registered IDs only
```

### Concrete changes

1. **Extension point** (shared backend package / rhaap-common node surface)  
   e.g. `ansibleSettingsCapabilityExtensionPoint` with `add({ id, label? })`.

2. **Owning module registers the resource type** (not APME-gated)  
   After collecting registrations, call `permissionsRegistry.addResourceType` with:
   - `ansible.settings.edit`
   - `FOR_CAPABILITY` whose `paramsSchema` is built from the **registered** capability IDs  
   - `getResources: ref => registeredIds.includes(ref) ? { capability: ref } : undefined`

3. **APME module only registers + authorizes its own id**  
   ```ts
   settingsCapabilities.add({ id: 'apme', label: 'APME Quality' });
   // middleware / RequirePermission: resourceRef: 'apme'
   ```
   Parameterize middleware: `createRequireSettingsManageMiddleware({ capability: 'apme' })` instead of hardcoding `'apme'` inside the helper.

4. **Do not pre-list `aap` / `general`** until those modules exist and register. Keeps the RBAC dropdown honest.

### Restart / config contract (explicit)

| Change | Effect after backend restart |
|--------|------------------------------|
| Enable APME (or other settings owner) in config | Capability appears in `FOR_CAPABILITY` enum + `getResources` accepts it |
| Disable that plugin/module | Capability disappears from metadata / enum for new policy editing; existing policies referencing a removed id should DENY safely via `getResources` → `undefined` |
| Add a new settings-owning module | It calls `addCapability(...)` at init; no change to a central hardcoded enum required |

No live polling of APME is needed (or desirable) — APME is the data plane, not the portal capability registry.

### Near-term on this PR (if extension point is follow-up)

- Derive zod from a single `ANSIBLE_SETTINGS_CAPABILITIES` source (avoid duplicated literals).
- Ship only `'apme'` until other owners land.
- Parameterize `resourceRef` in middleware / frontend.
- Document: new capabilities = register at module init + backend restart; stock RHDH cannot load a live options API into this dropdown without custom UI or switching to `z.string()` (worse UX / typo DENYs).

Happy to follow up with an API sketch / small RFC for the extension point if useful.

ganeshrn (approved):

PR #550, opened 2026-08-12
fix(apme): move settings permission rules for portal dynamic export

djdanielsson:
## Summary
- Remove the invalid `./permissions/rules` package export from `backstage-rhaap-common` (Backstage CLI rejects nested export mount points during `export-dynamic`)
- Move APME settings permission rules into `catalog-backend-module-apme`, where they are registered
- Unblocks portal-local `make dev` plugin export (`yarn build` on embedded `backstage-rhaap-common`)

## Test plan
- [x] `yarn tsc` passes in `ansible-backstage-plugins`
- [x] `yarn build` in `plugins/backstage-rhaap-common` succeeds
- [x] `make dev` in `automation-portal-local` completes plugin export and starts the portal stack


NilashishC (changes requested):
This should be resolvable by https://github.com/ansible/ansible-backstage-plugins/pull/553 without moving the permissions stuff to apme plugin. Please give it a try.

